PR #10, opened 2022-03-07
Create LICENSE

fkorotkov:


sogajeffrey:
@edigaryev @fkorotkov why was AGPL selected as the License? many companies cannot use tools with AGPL license. Any chance it can be changed to something like Apache or MIT?

fkorotkov:
Hey @sogajeffrey, coincidently someone asked the same question yesterday and I tried to explain our reasoning here https://github.com/cirruslabs/tart/discussions/369. Feel free to comment there or email me directly at fedor@cirruslabs.org.

oleg-z:
@fkorotkov no coincidence here. We're on the same team =)

fkorotkov:
😅 got it! I suspected that. Might be a good idea to have a quick call next week or after NYE to talk about your concerns and how can we help but at the same time be protected from AWS-like takeovers.

PR #48, opened 2022-05-06
Softnet: an alternative to built-in NAT with better isolation

### LINKED ISSUE #20: Investigate how to isolate two tart VMs
Seems by default Virtualization framework makes the network sharable. Let's see if we can isolate two VMs running simultaneously so one can't ssh into another.

edigaryev:
This is how it looks from the Tart's side, the `softnet` binary will be separately implemented in Rust.

Resolves #20.

### DIFF
--- a/Sources/tart/Commands/Run.swift
+++ b/Sources/tart/Commands/Run.swift
@@ -43,16 +43,7 @@ struct Run: AsyncParsableCommand {
     let task = Task {
       do {
         if let vncWrapper = vncWrapper {
-          do {
-            let (port, password) = try await vncWrapper.credentials()
-            let url = URL(string: "vnc://:\(password)@127.0.0.1:\(port)")!
-            print("Opening \(url)...")
-            if ProcessInfo.processInfo.environment["CI"] == nil {
-              NSWorkspace.shared.open(url)
-            }
-          } catch {
-            print("Failed to get an IP for screen sharing: \(error)")
-          }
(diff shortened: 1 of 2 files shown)


Sources/tart/Commands/Run.swift
```diff
@@ -27,24 +27,38 @@ struct Run: AsyncParsableCommand {
...
+            let (port, password) = try await vncWrapper.credentials()
+            let url = URL(string: "vnc://:\(password)@127.0.0.1:\(port)")!
+            print("Opening \(url)...")
+            if ProcessInfo.processInfo.environment["CI"] == nil {
+              NSWorkspace.shared.open(url)
+            }
+          } catch {
+            print("Failed to get an IP for screen sharing: \(error)")
+          }
+        }
```
fkorotkov, line 56: What about putting the opening logic in a `VNCWrapper` method and do something like the following:

```suggestion
        if vnc && ProcessInfo.processInfo.environment["CI"] == nil {
          vncWrapper?.openURL()
        }
```

fkorotkov:
Should we add softnet as a dependency to Goreleaser config as part of this PR or a separate?

fkorotkov (approved):

edigaryev:
> Should we add softnet as a dependency to Goreleaser config as part of this PR or a separate?

Let's add this separately once the Softnet is published both on GitHub and Homebrew.

PR #58, opened 2022-05-10
Two more FAQs

fkorotkov:
@mikehardy here is an answer from [your question](https://github.com/actions/virtual-environments/issues/2187#issuecomment-1122700339) about nested virtualization support. Unfortunately, it's not supported yet and you can't run something like Docker inside Tart VMs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,7 +162,7 @@ tart pull acme.io/remoteorg/name:latest my-local-vm-name
 </details>
 
 <details>
-  <summary>How to change disk size of a given VM?</summary>
+  <summary>How to change VM's disk size?</summary>
 
   You can choose disk size upon creation of a virtual machine:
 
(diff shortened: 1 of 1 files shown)


mikehardy:
Thanks! I really appreciate the response. Without nested virt my use case of app testing (where you need an android emulator, which implies hardware acceleration for the emulator, thus nested virt) isn't supported but I totally understand we can all only do what the tech lets us do. OSX-KVM is where I'll be for a while then. Maybe they'll get it all working on arm64 soon if it's not already :laughing:  Cheers!

fkorotkov:
@mikehardy yeah! I'm very looking forward to WWDC next month and and the new macOS with further improvement of the `Virtualization.Framwework`.

edigaryev (approved):

PR #64, opened 2022-05-11
.goreleaser(brews): depend on macOS Monterey

### LINKED ISSUE #63: Check macOS version is supported
running the first steps, i get the following error:

```
tart clone ghcr.io/cirruslabs/macos-monterey-base:latest monterey-base
dyld: Library not loaded: /usr/lib/swift/libswift_Concurrency.dylib
  Referenced from: /opt/homebrew/bin/tart
  Reason: image not found
zsh: abort      tart clone ghcr.io/cirruslabs/macos-monterey-base:latest monterey-base
```
M1, macOS 11.4

edigaryev:
Resolves #63.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -35,3 +35,5 @@ brews:
     homepage: https://github.com/cirruslabs/tart
     description: Run macOS VMs on Apple Silicon
     skip_upload: auto
+    custom_block: |
+      depends_on :macos => :monterey
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ Built by CI engineers for your automation needs. Here are some highlights of Tar
 * Use Tart Packer Plugin to automate VM creation.
 * Built-in CI integration.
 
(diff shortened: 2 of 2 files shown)


.goreleaser.yml
```diff
@@ -35,3 +35,5 @@ brews:
     homepage: https://github.com/cirruslabs/tart
     description: Run macOS VMs on Apple Silicon
     skip_upload: auto
+    custom_block: |
+      depends_on :macos => :monterey
```
fkorotkov, line 39: Is it just Monterey or Monterey and newer?

edigaryev: It seems to be the latter, can be validated with `depends_on :xcode => "15.0"` and the docs also say:

>Some [Requirement](https://rubydoc.brew.sh/Requirement)s can also take a string specifying their minimum version that the formula depends on.

(https://docs.brew.sh/Formula-Cookbook#specifying-other-formulae-as-dependencies)

fkorotkov (approved):

PR #65, opened 2022-05-11
Registry: use issued_at field in the token response message

edigaryev:
A follow-up to https://github.com/cirruslabs/tart/pull/55.

### DIFF
--- a/Sources/tart/OCI/Registry.swift
+++ b/Sources/tart/OCI/Registry.swift
@@ -8,11 +8,27 @@ enum RegistryError: Error {
 }
 
 struct TokenResponse: Decodable {
-  let creationTime = Date()
-  
   var token: String
-  var expires_in: Int?
-  
+  var expiresIn: Int?
+  var issuedAt: Date = Date()
+
+  enum CodingKeys: String, CodingKey {
+    case token
+    case expiresIn = "expires_in"
+    case issuedAt = "issued_at"
+  }
+
+  init(from decoder: Decoder) throws {
+    let container = try decoder.container(keyedBy: CodingKeys.self)
+
+    token = try container.decode(String.self, forKey: .token)
+    expiresIn = try container.decodeIfPresent(Int.self, forKey: .expiresIn)
+    if let issuedAtRaw = try container.decodeIfPresent(String.self, forKey: .issuedAt),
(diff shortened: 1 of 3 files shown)


Sources/tart/OCI/Registry.swift
```diff
@@ -8,11 +8,27 @@ enum RegistryError: Error {
...
 struct TokenResponse: Decodable {
-  let creationTime = Date()
-  
   var token: String
-  var expires_in: Int?
-  
+  var expiresIn: Int?
+  var issuedAt: Date = Date()
+
+  enum CodingKeys: String, CodingKey {
```
fkorotkov, line 15: I'm not a big fan of writing custom decoding that way. `TokenResponse` is too simple for that. Let's maybe use [`dateDecodingStrategy`](https://developer.apple.com/documentation/foundation/jsondecoder/2895216-datedecodingstrategy) of the `JSONDecoder`?

edigaryev: I don't think this is a good idea:

1. Using snake case might simplify things at first, but once we start using a code formatter that enforces [official Swift's coding conventions](https://www.swift.org/documentation/api-design-guidelines/#general-conventions) these fields case will be converted and token parsing will break suddenly
2. Leaking abstractions: to parse the `TokenResponse` correctly we will need a specifically-configured `JSONDecoder` instance, so naturally we may want to hide these details in a custom `TokenResponse.init()`, however from the outside view the `TokenResponse` will still be `Decodable`, and thus one might decode it using some other `JSONDecoder` instance accidentally
3. General awkwardness: since there's a `let` field in `TokenResponse`, we cannot simply assign `JSONDecoder().decode(...)` result to `self` in custom `init()`, we will need to assign each `var` field one by one

fkorotkov: 1. There is `decoder.keyDecodingStrategy = .convertFromSnakeCase`. 🤔 
2. Might use a static method like `TokenResponse.parse`.
3. I think it's less awkward then having the keys and custom decoding logic. We can make `TokenResponse` `fileprivate` so it's not leaking beside the file.

edigaryev: See 7a53744d5b7205b474c0cb908569e9d3adda5f1a.

fkorotkov (approved):

Issue #44, opened 2022-05-04
Option to run a VM with an empty attached disk


fkorotkov:
Right now if you want to install Xcode you need to create a large VM in order to download the `*.xip` and unarchive it. There is no easy option to unarchive inplace so it will be great if we can launch a VM with some empty disk that we can use as a temporary location for such disk hungry operations.

anjannath:
I think i've hit this issue, i wasn't able to install homebrew on a newly created vm with the default disk size, and looking at this issue it could be that installing the SDK tools (`xcode-select --install`) failed due to the limited disk size.

Can you please provide more details of how your proposed solution can be implemented

fkorotkov:
@anjannath not sure if that will help in your case since you can tell `xcode-select` where to download files to. In you case you might try to install `brew` which seems installing the command line tools in a better manner and it [worked for us with a 30GB disk](https://github.com/cirruslabs/macos-image-templates/blob/master/templates/base.pkr.hcl#L32) (note that Tar Packer plugin also removes recovery partition which saves around 5 GB). You can also specify a bigger disk (the default is 32 GB) for your image with `--disk-size` flag. For example, `tart create --from-ipsw=latest --disk-size=40 monterey-vanilla`.

For this issue specifically, I meant that right now [Tart only attaches one disk to the VM](https://github.com/cirruslabs/tart/blob/8554e56e4baa8b31431d69858282e05a125d22d2/Sources/tart/VM.swift#L186-L189). What we can do is to create a temporary disk file, format it and attache to the VM upon `run`. This way there will be kind of a buffer we can download `Xcode.xip` file, unarchive it into `Xcode.App` and copy only `Xcode.App` into `/Application` folder on the main persistent disk.

Lutzifer:
@fkorotkov Is copying Xcode to the guest via SCP a viable alternative?

fkorotkov:
Yeah, only you copy the unxiped `Xcode.App` itself. But I still believe an empty attached disk might be helpful for cases of CI where a CI build can clone huge repositories into the attached disk and run the build in it.

This ticket was created from the idea to not require to create huge 100 GB VMs in order to fit all the workloads but have an option to expand the available disk space on demand.

fkorotkov:
I guess we can close this issue in favor of #13 which will be even a better option for just mounting a host folder.

PR #71, opened 2022-05-15
Change default disk size upon creation

fkorotkov:
To help with frustrations like https://github.com/cirruslabs/tart/issues/44#issuecomment-1126831261

Experts will still use `--disk-size` flag to tailor disk size for their needs.


fkorotkov:
I think with this we'll cut `0.4.0`.

edigaryev (approved):

Issue #75, opened 2022-05-16
tart list shows files like .DS_Store as VMs


Lutzifer:
if there is a file in the vms folder (e.g. .DS_Store) it is shown as a VM.

```
➜ arm64  vms pwd                                                                                            [22/05/16| 1:27PM]
/Users/admin/.tart/vms
➜ arm64  vms ls -al                                                                                         [22/05/16| 1:27PM]
total 0
drwxr-xr-x  4 admin  staff  128 May 16 13:26 .
drwxr-xr-x  5 admin  staff  160 May 16 11:49 ..
-rw-r--r--  1 admin  staff    0 May 16 13:26 .local_file
drwxr-xr-x  5 admin  staff  160 May 16 09:58 monterey-vanilla
➜ arm64  vms tart list                                                                                      [22/05/16| 1:27PM]
Name	Source
local	monterey-vanilla
local	.local_file
```

PR #76, opened 2022-05-16
tart list: filter out uninitialized VM directories

edigaryev:
Similarly to what's done in `VMStorageOCI`: https://github.com/cirruslabs/tart/blob/efc9c74b6c557862fad8f5d1316976d3f772eb33/Sources/tart/VMStorageOCI.swift#L45-L47

Resolves https://github.com/cirruslabs/tart/issues/75.


Lutzifer:
running this using

```
mint run cirruslabs/tart@filter-uninitialized list  
```

fixes #75 for me.

Thanks for the ultra fast response!

PR #78, opened 2022-05-16
Add virtual sound devices

fkorotkov:
I tried it once but didn't provide `sources` and then the VM just hanged so I assumed Virtio sound devices doesn't work similar to folder sharing. But I tried again and it worked!

### DIFF
--- a/Sources/tart/VM.swift
+++ b/Sources/tart/VM.swift
@@ -173,6 +173,10 @@ class VM: NSObject, VZVirtualMachineDelegate, ObservableObject {
     ]
     configuration.graphicsDevices = [graphicsDeviceConfiguration]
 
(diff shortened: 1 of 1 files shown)


Sources/tart/VM.swift
```diff
@@ -173,6 +173,10 @@ class VM: NSObject, VZVirtualMachineDelegate, ObservableObject {
     ]
     configuration.graphicsDevices = [graphicsDeviceConfiguration]
 
+    let soundDeviceConfiguration = VZVirtioSoundDeviceConfiguration()
```
edigaryev, line 176:
```suggestion
    // Audio
    let soundDeviceConfiguration = VZVirtioSoundDeviceConfiguration()
```

Issue #14, opened 2022-03-28
Support pasting clipboard from host


fkorotkov:
It seems it's not supported at the moment with the current API according to [this comment](https://github.com/KhaosT/MacVM/issues/14) and based on my research at the time. But let's create a tracking issue.

fkorotkov:
It seems it will be possible. At least according to a [new Linux example](https://developer.apple.com/documentation/virtualization/running_gui_linux_in_a_virtual_machine_on_a_mac) (see "Enable copy-and-paste support between the host and the guest" section):

```swift
private func createSpiceAgentConsoleDeviceConfiguration() -> VZVirtioConsoleDeviceConfiguration {
    let consoleDevice = VZVirtioConsoleDeviceConfiguration()

    let spiceAgentPort = VZVirtioConsolePortConfiguration()
    spiceAgentPort.name = VZSpiceAgentPortAttachment.spiceAgentPortName
    spiceAgentPort.attachment = VZSpiceAgentPortAttachment()
    consoleDevice.ports[0] = spiceAgentPort

    return consoleDevice
}
```

Need to verify that it will work in macOS though.

fkorotkov:
I've tried the above example and a bit different one:

```swift
let spiceAgentPortAttachment = VZSpiceAgentPortAttachment()
spiceAgentPortAttachment.sharesClipboard = true
let serialPortConfiguration = VZVirtioConsoleDeviceSerialPortConfiguration()
serialPortConfiguration.attachment = spiceAgentPortAttachment
configuration.serialPorts = [serialPortConfiguration]
```

with the latest Ventura and it seems not be working. It might be indicating that it only works for Linux. Will need to investigate more.

bootstraponline:
@fkorotkov Hey, https://github.com/cirruslabs/tart/commit/048a5506df871aaddc412ddfabe2e52bec850e57 doesn't support clipboard sharing on Linux. A spice agent is necessary, similar to what you posted in https://github.com/cirruslabs/tart/issues/14#issuecomment-1147850768

https://github.com/cirruslabs/tart/pull/287 also fixes this issue. Would you prefer a standalone PR for clipboard sharing?

fkorotkov:
[...]

PR #97, opened 2022-05-20
Option to run with VNC

fkorotkov:
So copy/paste is working. For example, `tart run --vnc latest`.

Related to #14

### DIFF
--- a/Sources/tart/Commands/Run.swift
+++ b/Sources/tart/Commands/Run.swift
@@ -34,6 +34,7 @@ struct Run: AsyncParsableCommand {
       if vnc {
         group.addTask(operation: {
           do {
+            print("Waiting for the VM to boot...")
             let resolvedIP = try await IP.resolveIP(vm!.config, secondsToWait: 60)
             guard let ip = resolvedIP else {
               throw IPNotFound()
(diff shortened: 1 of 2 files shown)


Sources/tart/Commands/Run.swift
```diff
@@ -14,13 +17,35 @@ struct Run: AsyncParsableCommand {
   @Flag var noGraphics: Bool = false
 
   @Flag var recovery: Bool = false
+  
+  @Flag var vnc: Bool = false
```
edigaryev, line 21: It would be nice to add an explanation on what this flag does, because someone might think that it configures and starts a VNC server automatically for them, which is not true.

edigaryev, line 21: It would be nice to add an explanation on what this flag does, because someone might think that it configures and starts a VNC server automatically for them, which is not true.

edigaryev:
What do you think about adding an additional `tart vnc` command, which would allow opening multiple sessions to the VM and keeping the remote control available in case the original session was lost for some reason?

edigaryev (approved):

Sources/tart/Commands/Run.swift
fkorotkov: You were too quick. Opened #100 with the followup on this

PR #98, opened 2022-05-20
Fixed build warnings

fkorotkov:
`swift build` is not printing any warnings now

### DIFF
--- a/Sources/tart/OCI/Registry.swift
+++ b/Sources/tart/OCI/Registry.swift
@@ -26,8 +26,8 @@ extension HTTPClientResponse.Body {
 }
 
 struct TokenResponse: Decodable {
-  static let defaultIssuedAt = Date()
-  static let defaultExpiresIn = 60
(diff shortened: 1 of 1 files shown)


Sources/tart/OCI/Registry.swift
```diff
@@ -26,8 +26,8 @@ extension HTTPClientResponse.Body {
 }
 
 struct TokenResponse: Decodable {
-  let defaultIssuedAt = Date()
-  let defaultExpiresIn = 60
+  private static let defaultIssuedAt = Date()
```
edigaryev, line 29: I don't think it's possible for this to be static: the actual `Date` needs to be filled when we get the token, not when we initialize the global variables.

fkorotkov: Right! My bad was copy/pasting without thinking. Found https://github.com/marksands/BetterCodable. Maybe will rework all codables that way to also avoid non default date parsing strategy. Will close OR for now

Issue #39, opened 2022-05-03
Pushing of an image should populate local cache


fkorotkov:
Once one successfully pushes an image via:

```shell
tart push my-local some-cr.io/acme/remote:latest
```

Tart should copy over `my-local` into `oci` cache.

erikng:
What about a symlink?

edigaryev:
> What about a symlink?

Both symbolic link and the hard link won't ensure that the local VM's disk and configuration will stay immutable. Locking a pushed local VM from running and making changes to it is probably not an option.

fkorotkov:
@edigaryev what do you think of a flag like `--populate-cache`  which will be `false` by default?

PR #103, opened 2022-05-24
tart pull: introduce --populate-cache flag

edigaryev:
Resolves https://github.com/cirruslabs/tart/issues/39.

### DIFF
--- a/Sources/tart/Commands/Push.swift
+++ b/Sources/tart/Commands/Push.swift
@@ -12,6 +12,10 @@ struct Push: AsyncParsableCommand {
   @Argument(help: "remote VM name(s)")
   var remoteNames: [String]
 
+  @Flag(help: ArgumentHelp("cache pushed images locally",
+          discussion: "Increases disk usage, but saves time if you're going to pull the pushed images later."))
+  var populateCache: Bool = false
+
   func run() async throws {
     do {
       let localVMDir = try VMStorageLocal().open(localName)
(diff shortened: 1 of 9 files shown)


Sources/tart/Commands/Push.swift
```diff
@@ -32,15 +36,41 @@ struct Push: AsyncParsableCommand {
...
+        }
+      }
+
+      // Populate local cache (if requested)
+      if populateCache {
+        for pushedToRemoteName in pushedToRemoteNames {
+          defaultLogger.appendNewLine("caching \(localName) as \(pushedToRemoteName)...")
+
+          let ociVMDir = try VMStorageOCI().create(pushedToRemoteName, overwrite: true)
+          try localVMDir.clone(to: ociVMDir, generateMAC: false)
```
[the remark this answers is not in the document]
edigaryev: Good catch! See https://github.com/cirruslabs/tart/pull/103/commits/6cc5402faa314250a9f3976eeea9c1b0c98aba67.

Sources/tart/Commands/Push.swift
```diff
@@ -35,12 +39,21 @@ struct Push: AsyncParsableCommand {
...
+        // Populate the local cache (if requested)
+        if populateCache {
+          for remoteName in remoteNamesForRegistry {
+            defaultLogger.appendNewLine("caching \(localName) as \(remoteName)...")
+
+            if let cacheToVMDir = try VMStorageOCI().cache(name: remoteName, digest: manifestDigest) {
+              try localVMDir.clone(to: cacheToVMDir, generateMAC: false)
+            }
+          }
+        }
```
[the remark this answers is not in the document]
edigaryev: >And I still think there is too much cloning is going on.

It only clones if the "digest" directory doesn't exist, otherwise creates a symbolic link. This is achieved by re-using the `pull()` code.

> What if `pushToRegistry` will return a `RemoteName with the `shaXXX` then the cache population logic might be much simpler:

I see two downsides to the example you've showed:

1. It seems that the `localVMDir.clone()` will always clone.
2. Not reusing the `pull()`'s logic results in needless duplication.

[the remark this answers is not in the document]
fkorotkov: I don't think there will much "needless duplication". 🤷‍♂️ Here is my take #107

Sources/tart/Commands/SetCommand.swift
```diff
@@ -1,8 +1,8 @@
 import ArgumentParser
 import Foundation
 
-struct Set: AsyncParsableCommand {
-  static var configuration = CommandConfiguration(abstract: "Modify VM's configuration")
+struct SetCommand: AsyncParsableCommand {
```
fkorotkov, line 4: All the files in `Commands` folder doesn't have `Command` suffix. What's the reasoning for this particular rename?

edigaryev: > What's the reasoning for this particular rename?

`Set` command collides with the [`Set`](https://developer.apple.com/documentation/swift/set) collection.

fkorotkov: I don't think it's in the scope of this PR. There is no usage of Set. IMO we shouild either rename all the commands or wait util there will be a collision.

edigaryev: Reverted in https://github.com/cirruslabs/tart/pull/103/commits/d59f4ec06686fa55d0e3c5e151980e59535990e5.

fkorotkov (approved):

PR #106, opened 2022-05-25
Auto-detect screen DPI 

### LINKED ISSUE #104: Detect screen DPI automatically
If I compare to my real machine, or to another implementation of Apple's virtualisation framework, the display is very blurry/fuzzy in tart VMs.

Real (top), Tart (centre), Other (bottom)

_Real (top), Tart (centre), Other (bottom)_

fkorotkov:
It appeared there is a hand `(for screen: NSScreen, sizeInPoints: NSSize)` constructor of `VZMacGraphicsDisplayConfiguration` which does the configuration for us. So since a new windows always opens on the main screen, let's use this constructor in case there is a main screen. Otherwise we'll do a reasonable guess of the DPI.

Fixes #104

### DIFF
--- a/Sources/tart/Commands/Set.swift
+++ b/Sources/tart/Commands/Set.swift
@@ -13,7 +13,7 @@ struct Set: AsyncParsableCommand {
   @Option(help: "VM memory size in megabytes")
   var memory: UInt16?
 
-  @Option(help: "VM display settings in a format of <width>x<height>(x<dpi>)?. For example, 1200x800 or 1200x800x72")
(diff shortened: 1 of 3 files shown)


edigaryev (approved):

ryanbooker:
What does "main screen" mean in this context? When I launch a vm from the command line the window opens on the same screen as the terminal session.

fkorotkov:
According to the [docs](https://developer.apple.com/documentation/appkit/nsscreen/1388371-mainscreen) it's:

> the screen object containing the window with the keyboard focus.

So I think it will be OK.

PR #107, opened 2022-05-27
Review comments for #103

fkorotkov:


Sources/tart/Commands/Push.swift
```diff
@@ -42,16 +42,15 @@ struct Push: AsyncParsableCommand {
...
           + "\(registryIdentifier.host)/\(registryIdentifier.namespace)\(remoteNamesForRegistry.referenceNames())...")
 
-        let manifestDigest = try await localVMDir.pushToRegistry(registry: registry, references: remoteNamesForRegistry.map{ $0.reference.value })
+        let pushedRemoteName = try await localVMDir.pushToRegistry(registry: registry, references: remoteNamesForRegistry.map{ $0.reference.value })
 
         // Populate the local cache (if requested)
         if populateCache {
+          let ociStorage = VMStorageOCI()
+          let expectedPushedVMDir = try ociStorage.create(pushedRemoteName)
+          try localVMDir.clone(to: expectedPushedVMDir, generateMAC: false)
```
edigaryev, line 51: Looks like this will overwrite all the VM's files, including disk, even if an identical VM with the same digest already exists?

fkorotkov: I think it's practically impossible to get the same digest. Isn't it? I thought it's a hash of a newly formed manifest files from the uploaded layers.

edigaryev: > I think it's practically impossible to get the same digest. Isn't it? I thought it's a hash of a newly formed manifest files from the uploaded layers.

I think it quite possible if you push the same VM multiple times (e.g. to different registries).

It's tempting to neglect this, but on the other hand I'm not sure if overwriting 70 GB files is good for SSD and it seems that it's quite easy for us to skip doing needless work just by calling `exists()`.

fkorotkov: Needless is a strong word in this context IMO. There are slight similarities but I don't see how it's identical. My concern was that these lines were really hard to understand:

```swift
let manifestDigest = try await localVMDir.pushToRegistry(registry: registry, references: remoteNamesForRegistry.map{ $0.reference.value })

// Populate the local cache (if requested)
if populateCache {
  for remoteName in remoteNamesForRegistry {
    defaultLogger.appendNewLine("caching \(localName) as \(remoteName)...")

    if let cacheToVMDir = try VMStorageOCI().cache(name: remoteName, digest: manifestDigest) {
      try localVMDir.clone(to: cacheToVMDir, generateMAC: false)
    }
  }
}
```

IMO it's not clear at all what is going on there without going into implementation details.

Issue #21, opened 2022-04-11
File based lock on VM operations


fkorotkov:
Let's make sure that a user can't access VM from two simultaneous `tart` invocations. We can probably just put a lock file inside the VMs folder. Plus we can modify timestamp of the file upon a lock so we potentially can see when the VM was used last time. Might be useful for auto-pruning of VMs based on LRU concept.

edigaryev:
I've started implementing this and realized that in the current state, in which there's no single daemon like `dockerd`, but instead multiple `tart` instances with the same access level that can run concurrently, the locking would create more problems that it'd solve.

Specifically, the complexities it'll bring to the codebase may outweigh the damage caused by issuing `tart delete` on a running instance.

What I propose at this point instead is to perform all writes consisting of multiple files into a temporary directory and then `rename(2)` these into their final destination. This will ensure that `tart create`, `tart clone` and `tart pull` would avoid inconsistent state.

@fkorotkov what do you think?

fkorotkov:
Agreed that it can bring unnecessary complexity. I think the minimum we should do is to:
1. Clean up the vm directory in case of failure or Ctrl+C for `create`/`pull`. 
2. `clone` is already an atomic file system operation.
3. Print a nice error in case someone tries to `run` the same VM twice like `Vitual machine "latest" is already running with 192.168.65.124 IP`.

One other note is to maybe think of adding a file to the VM directory timestamp of which will represent time of the last operation with the VM. But something like that should be a part of #37 (we discussed that the lock files from this issue could play a role of the timestamp data).

PR #109, opened 2022-05-31
Atomic tart {create,clone,pull} operations using rename(2)

edigaryev:
Assuming that `FileManager`'s `replaceItemAt()` [has `rename(2)` semantics](https://rcrowley.org/2010/01/06/things-unix-can-do-atomically.html).

Resolves https://github.com/cirruslabs/tart/issues/21.

### DIFF
--- a/Sources/tart/Root.swift
+++ b/Sources/tart/Root.swift
@@ -20,11 +20,15 @@ struct Root: AsyncParsableCommand {
     ])
 
   public static func main() async throws {
-    // Handle cancellation by Ctrl+C
+    // Ensure the default SIGINT handled is disabled,
+    // otherwise there's a race between two handlers
+    signal(SIGINT, SIG_IGN);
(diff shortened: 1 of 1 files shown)


Sources/tart/Commands/Create.swift
```diff
@@ -23,14 +23,16 @@ struct Create: AsyncParsableCommand {
 
   func run() async throws {
     do {
-      let vmDir = try VMStorageLocal().create(name)
+      let tmpVMDir = try VMDirectory.temporary()
```
fkorotkov, line 26: will it be cleaned up in case of a Ctrl+C?

edigaryev: Now it will be: https://github.com/cirruslabs/tart/pull/109/commits/4037f7ffb1e562c38920f0e12c8ab5e3355126d1.

fkorotkov (approved):

PR #122, opened 2022-06-09
Accept disk size > 255GB, fixing issues on Ventura

keisau:
* `UInt8` -> `Int` for supporting larger disk size
* On Ventura, `dispatchMain` inside `withThrowingTaskGroup` results in unexpected behaviors:
    - `Segmentation Fault` for `--no-graphics`
    - Screen shows up but the virtual machine doesn't show up otherwise


Sources/tart/Commands/Run.swift
```diff
@@ -69,12 +69,12 @@ struct Run: AsyncParsableCommand {
...
-      } else {
-        runUI()
-      }
+    }
+    
+    if noGraphics || vnc {
+      dispatchMain()
+    } else {
+      runUI()
     }
```
fkorotkov, line 78: Could you please check if #123 works for you? With this change it seems UI won't be shown at all because it will wait for the group to finish aka VM stops.

keisau: Ventura was too problematic to me, restored to Monterey now. Will find a M1 machine to test that. Withdrawn the commit and leave it for now.

Sources/tart/Commands/Create.swift
```diff
@@ -13,7 +13,7 @@ struct Create: AsyncParsableCommand {
   var fromIPSW: String?
 
   @Option(help: ArgumentHelp("Disk size in Gb")) 
-  var diskSize: UInt8 = 50
+  var diskSize: Int = 50
```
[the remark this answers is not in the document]
keisau: Good point. Updated the PR accordingly.

Sources/tart/Commands/Run.swift
keisau: Just tried #123 , works except Windowed mode

fkorotkov (approved):

fkorotkov:
Hey @keisau,

We are preparing 1.0.0 release of Tart at some point in January and would like to clean things up and stabilize the API to provide long term support. Therefore we are introducing Contributor License Agreement which is common practice among mature OSS projects.

PR #123, opened 2022-06-09
Don't use dispatchMain

fkorotkov:
Seems it's crashing on Ventura as reported in #122

### DIFF
--- a/Sources/tart/Commands/Run.swift
+++ b/Sources/tart/Commands/Run.swift
@@ -37,23 +37,7 @@ struct Run: AsyncParsableCommand {
     let vmDir = try VMStorageLocal().open(name)
     vm = try VM(vmDir: vmDir)
 
-    await withThrowingTaskGroup(of: Void.self) { group in
-      if vnc {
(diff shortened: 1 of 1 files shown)


edigaryev (approved):

keisau:
On Ventura, `--no-graphics` mode works. VNC seems to work as well (I don't have a VNC enabled image), but the VM isn't starting in Windowed mode, eg: `tart run macos` (without any additional arguments)

fkorotkov:
I might enroll a device next week so I'll have a non-virtualized Ventura host to reproduce and debug. 

In the meanwhile, @keisau could you please also check `system.log` in `Console.App` to see if there are any logs from `Virtualization.Framework`?

keisau:
Ventura is just so unstable that I restored back to Monterey again. Will have another M1 (or M2) to get you the logs. But I have to wait......

fkorotkov:
Let's merge the change since it seems like an improvement. Will test on Ventura later this week.

Issue #101, opened 2022-05-23
Automate creation from scratch


fkorotkov:
Right now when one creates a VM from scratch there is a need for manual input from a user e.g. click thought the macOS installation process. It will be great if we can automate the process and do the [recommended steps](https://github.com/cirruslabs/tart#creating-from-scratch) automatically.

```console
tart create --from-ipsw=latest --username=admin --password=admin latest
```

This way we can also start saving the username and password within the config to simplify UX for cases like `tart run --vnc latest` for example (can pre-fill username and password for screen sharing).

trodemaster:
Building custom macOS VMs starting with a local ipsw file is my goal as well. See my existing packer template for x86 macOS VMs here https://github.com/trodemaster/packer-macOS-11

My approach would be similar for Applesilcon VMs. Which is to boot the VM and then use packer boot commands to complete the Apple Setup screens and ssh configuration so that packer could finish the customizations.

I'm unclear on what the username and password would do with the tart cli tool? Is that maybe overlapping with the packer boot commands function? Keep in mind apple likes to change those setup screens with major releases so could get messy if you bake that directly into tart.

damonbreeden:
related: https://github.com/cirruslabs/macos-image-templates/pull/48

trodemaster:
Using an MDM solution to push a profile to the VM Is an option.  I opened an issue with apple (FB10026741) asking them to add support for simply passing a profile to the VM via the virtualization framework directly bypassing the networking and MDM requirements. I'm not holding my breath on that one. But if others request the same thing it would help. 

Straight up building https://micromdm.io into tart would be a robust but likely overly complex solution. Tart and its packer plugin are really only missing a handful of features to enable building from scratch via VNC commands and existing techniques.

fkorotkov:
@trodemaster agreed that your technique seems like the best option. We just haven't had time to look into your packer plugin. I briefly checked out the code but I'm not too familiar with the things there so need definitely a more thorough look 😅  Hopefully will find time next week.

trodemaster:
Sounds good! I have posted a screen recording of the process on my blog which should help visualize what's happening with the boot command technique. It's the second video on this page. https://netjibbing.com/post/packer-macos-11/  Applesilicon macs boot directly into the setup screens so in that case, the boot commands would need to work their way through setup. Not ideal but doable.

torarnv:
[...]

PR #126, opened 2022-06-14
Full-fledged VNC support

edigaryev:
Might be handy in https://github.com/cirruslabs/tart/issues/101.


Sources/tart/VNCWrapper.swift
```diff
@@ -0,0 +1,47 @@
+import Foundation
+import Dynamic
+import Virtualization
+
+class VNCWrapper {
+    private let vnc: Dynamic
+    private let vncCredentialsURL: URL
+
+    init(virtualMachine: VZVirtualMachine, vncCredentialsURL: URL) async throws {
+        let password = UUID().uuidString
```
fkorotkov, line 10: Maybe just use `admin` password? Or something like `password101` 😅

fkorotkov (commented):
I'm not a fan of the idea of storing VNC credentials. Plus `tart vnc` command is weird (see my comment in [VNC.swift](https://github.com/cirruslabs/tart/pull/126/files#diff-bff056509c8cbec825b7940a73826f12cb2176d72331dd9e6e3a29aece889fe9).

I think it will be easier to scope the change only to the `tart run --vnc` flag and not introduce credentials storing and a separate `vnc` command.

Sources/tart/VNCWrapper.swift
edigaryev: Can you explain your use-case a little better? Because the `tart vnc --open` relieves the paint of typing or copy-paste'ing that by hand.

fkorotkov: `tart vnc` is relying on the credentials being stored aka VM started with `tart run --vnc`.

But my comment was mostly for the UX since we'll print the `vnc://...` and it will be nice IMO to not have long UUID there.

Sources/tart/Commands/VNC.swift
```diff
@@ -0,0 +1,64 @@
+import ArgumentParser
+import Foundation
+import Network
+import SystemConfiguration
+
+struct VNCCredentialsNotFound: Error {
+}
+
+struct VNC: AsyncParsableCommand {
```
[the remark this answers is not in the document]
edigaryev: > This command will only work if a VM was started via `run --vnc` and we saved the credentials.

This PR changes things a little and makes the VNC server always available because it's cheap now.

Sources/tart/VNCWrapper.swift
edigaryev: > But my comment was mostly for the UX since we'll print the `vnc://...` and it will be nice IMO to not have long UUID there.

Would random human-readable passphrase work?

Sources/tart/Commands/VNC.swift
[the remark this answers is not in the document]
edigaryev: So, should it be completely disabled by default then? And only enabled with something like `--vnc-plus`? While the `--vnc` would use the mechanism you've mentioned above:

> Note that you can still VNC into a VM if _Remote Login_ option is enabled inside the VM.

Sources/tart/VNCWrapper.swift
fkorotkov: SGTM

Sources/tart/Commands/VNC.swift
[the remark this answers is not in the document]
fkorotkov: Initially I though of just replacing the existing logic with the private API in the `if vnc` statement.

Sources/tart/VNCWrapper.swift
edigaryev: See https://github.com/cirruslabs/tart/pull/126/commits/86563410962f2b1040a8db93ade15d7344004dee.

Sources/tart/Commands/VNC.swift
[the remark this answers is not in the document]
edigaryev: Implemented in https://github.com/cirruslabs/tart/pull/126/commits/f5a491111bcf05fceb604f2f3e1c1e076cd6e381 and  https://github.com/cirruslabs/tart/pull/126/commits/ff27894b0afff93f032c3e6270f125af79d3ae3f.

Sources/tart/Commands/VNC.swift
```diff
@@ -0,0 +1,89 @@
...
+                if open {
+                    vncCredentials.open()
+                } else {
+                    print(vncCredentials.url())
+                }
+
+                Foundation.exit(0)
+            }
+
+            print("no VNC credentials found, is your VM running?")
```
fkorotkov, line 48: I think this phrase is confusing if I just did `tart run macos` and then `open vnc:/<ip>`. What kind of credential are we talking about here?

Having a separate `vnc` command seems just make things more complex. We initially talked only about using the private APIs instead of the current implementation of `--vnc` flag for `run` command. Just to be able to VNC into a  VM crated from scratch.

edigaryev: > Having a separate `vnc` command seems just make things more complex. We initially talked only about using the private APIs instead of the current implementation of `--vnc` flag for `run` command. Just to be able to VNC into a VM crated from scratch.

Isn't the whole point of using the private API is the ability to do automatic installs via Packer? Which in turn requires a `tart vnc` command to be able to query the VNC credentials programatically and connect to the VNC using them?

Sources/tart/VM.swift
```diff
@@ -151,6 +155,10 @@ class VM: NSObject, VZVirtualMachineDelegate, ObservableObject {
       sema.signal()
     })
 
+    if let vncWrapper = vncWrapper {
+      try vncWrapper.stop()
```
fkorotkov, line 159: Maybe put it in `defer { ... }` in the if statement where VNC is started? 🤔

Sources/tart/Commands/Run.swift
```diff
@@ -37,9 +37,15 @@ struct Run: AsyncParsableCommand {
...
+    var vncWrapper: VNCWrapper?
+
+    if vnc {
+      vncWrapper = VNCWrapper()
+    }
+
     Task {
       do {
-        try await vm!.run(recovery)
+        try await vm!.run(recovery, vncWrapper)
```
fkorotkov, line 48: Not sure of `VM` should "know" about VNC. Seems `VNCWrapper` only needs an instance of `VZVirtualMachine`... So `VNCWrapper` can potentially be initialized with `vm!.virtualMachine` and do something like

```swift
vncWrapper.start()
defer { vncWrapper.stop }
try await vm!.run(recovery)
```

fkorotkov (commented):
Left some style/refactoring comments to discuss.

Sources/tart/Commands/Run.swift
edigaryev: See https://github.com/cirruslabs/tart/pull/126/commits/b07cf7711f59045a1817f8c6368061ca2643e05e.

Sources/tart/Commands/Run.swift
```diff
@@ -71,6 +75,10 @@ struct Run: AsyncParsableCommand {
     } else if !noGraphics {
       runUI()
     }
+
+    if let vncWrapper = vncWrapper {
```
fkorotkov, line 79: Could you please rebase after #132 to make sure the VNC will be stopped after the VM is stopped and not right away.

edigaryev: Done.

fkorotkov (approved):

PR #129, opened 2022-06-16
Fixes for Ventura

fkorotkov:
Still a noob in SwiftUI and Swift concurrency, but it seems on Ventura a task group is not actually running on main or something. Either way I think this change simplifies things but launching a VM in a task and then just continuing with either VNC or built-in graphics.

### DIFF
--- a/Sources/tart/Commands/Run.swift
+++ b/Sources/tart/Commands/Run.swift
@@ -31,48 +31,43 @@ struct Run: AsyncParsableCommand {
   func run() async throws {
     if recovery && vnc {
       print("You can't run in recovery and use VNC!")
-      Foundation.exit(1)      
+      Foundation.exit(1)
     }
     
     let vmDir = try VMStorageLocal().open(name)
     vm = try VM(vmDir: vmDir)
 
-    try await withThrowingTaskGroup(of: Void.self) { group in
-      group.addTask {
-        do {
-          try await vm!.run(recovery)
+    Task {
+      do {
+        try await vm!.run(recovery)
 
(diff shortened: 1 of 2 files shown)


keisau:
@fkorotkov I still don't have luck on my secondary M1 machine, but tested your branch on Monterey, work as usual!

Sources/tart/Commands/Run.swift
```diff
@@ -31,48 +31,43 @@ struct Run: AsyncParsableCommand {
...
+        }
+        let url = URL(string: "vnc://\(ip)")!
+        print("Opening \(url)")
+        NSWorkspace.shared.open(url)
+      } catch {
+        print("Failed to get an IP for screen sharing: \(error)")
       }
-
-      try await group.waitForAll()
+      vm?.wait()
```
edigaryev, line 68: Should this be a `vm!` instead (see `vm!.run()` above)?

And it seems that waiting for the `VM.sema` will result in premature termination:

https://github.com/cirruslabs/tart/blob/0d633de04af2bb454d7296f358ff51eafc77b559/Sources/tart/VM.swift#L139-L150

We probably need another semaphore that gets unlocked after the cleanup is done.

fkorotkov: Hot sure if that will corrupt anything but how about [b98b46d](https://github.com/cirruslabs/tart/pull/129/commits/b98b46d471bbea1075ad0cf8c29a7b28fb4f1724) then?

edigaryev (approved):

PR #137, opened 2022-06-24
Handle `tart run --no-graphics --vnc` properly

fkorotkov:
Let's start a VNC server but not force open Screen Sharing if `--no-graphics` is also passed.

### DIFF
--- a/Sources/tart/Commands/Run.swift
+++ b/Sources/tart/Commands/Run.swift
@@ -43,7 +43,16 @@ struct Run: AsyncParsableCommand {
     let task = Task {
       do {
         if let vncWrapper = vncWrapper {
-          await vncWrapper.open()
+          let port = try await vncWrapper.waitForPort()
+
+          let url = URL(string: "vnc://:\(vncWrapper.password)@127.0.0.1:\(port)")!
+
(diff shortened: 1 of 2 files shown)


Sources/tart/Commands/Run.swift
```diff
@@ -43,7 +43,16 @@ struct Run: AsyncParsableCommand {
     let task = Task {
       do {
         if let vncWrapper = vncWrapper {
-          await vncWrapper.open()
```
edigaryev, line 46: Maybe we can keep the opening logic in the `VNCWrapper` (see https://github.com/cirruslabs/tart/pull/48#discussion_r902768135) and pass a boolean that instructs the method whether to open the URL instead?

Something like:

```swift
await vncWrapper.wait(open: !noGraphics && ProcessInfo.processInfo.environment["CI"] == nil)
```

fkorotkov: I though about it but then it looks weird to me. Plus the `!noGraphics && ProcessInfo.processInfo.environment["CI"] == nil` expression looks a bit hard to read.

edigaryev (approved):

Issue #158, opened 2022-07-28
image push speed


victor-yousician:
I'm testing pushing an image (monterey-base from your documentation) to AWS ECR. I'm pushing from a Mac mini (M1, 2020) hosted in a data centre with a gigabit connection but only averaging 1MB/s. tart keeps one core at 100%. This seems pretty slow.

Nezz:
After 3 hours and 34 minutes we got to a whopping 44% of the image getting pushed:
```
time tart push monterey-base <redacted>.dkr.ecr.eu-west-1.amazonaws.com/ci/tart:20220728
pushing monterey-base to <redacted>.dkr.ecr.eu-west-1.amazonaws.com/ci/tart:20220728...
pushing disk... this will take a while...
44%
^CHTTPClientError.cancelled
tart push monterey-base   3756.16s user 9077.07s system 99% cpu 3:34:24.27 total
```

fkorotkov:
In our case we push to GitHub's Container Registry and we've seen speed of 30-40MB/s from three different locations in US all with a symmetric gigabit connection: two datacenters and my home. We attributed it to just how uploads work e.g. registries are optimized for pulls. 

But a 1 MB/s speed seems like unreasonably slow. How far your DC with the Mac Mini and AWS region?

edigaryev:
The 100% core usage seems weird.

Does it happen straight away after initiating the pull, or after some time?

Does it occur with other registries (e.g. when using GitHub Packages or Docker Hub) too?

victor-yousician:
> Does it happen straight away after initiating the pull, or after some time?
> Does it occur with other registries (e.g. when using GitHub Packages or Docker Hub) too?

This issue is with push speed, not pull. CPU goes to 100% (one core) immediately. The only other test I have is a Docker Registry container running locally. I don't think it's a meaningful data point, but CPU stays around 20-25% and speeds are around 35-45MB/s.

PR #159, opened 2022-07-28
Support chunked uploads

fkorotkov:
As described in https://docs.docker.com/registry/spec/api/#pushing-a-layer

Should help with #158

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -5,8 +5,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/swift-server/async-http-client",
       "state" : {
-        "revision" : "794dc9d42720af97cedd395e8cd2add9173ffd9a",
-        "version" : "1.11.1"
+        "revision" : "df87a860fdc41a595d5ca67f74cde9adbccc099a",
+        "version" : "1.11.4"
       }
     },
     {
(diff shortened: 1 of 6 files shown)


fkorotkov:
I've tested by uploading an image to ECR. Before the change it was barely doing anything, with the change I see 30-40 MB/s uploads. Will verify with GitHub Packages shortly too...

fkorotkov:
Just tried to push the latest Ventura beta to ghcr.io using the branch. Had to lower the chunk size and also fixed range end index (it's inclusive). See [6e3df64](https://github.com/cirruslabs/tart/pull/159/commits/6e3df64a7c98c9e618fa4727ee20caf85e2cb471)

fkorotkov:
But now ECR is not working with a 500 with no message. Seems the empty `PUT` is not working as expected. Will refactor to do the last put for the last chunk.

fkorotkov:
It appeared that not all registries support chunked uploads so I introduced a new `--chunk-size` option. 

`tart push --chunk-size=10` should fix #158 for @victor-yousician

fkorotkov:
Closing in favor of #168

fkorotkov:
Reopened for further discussion of this change being a feature rather then a bug fix.

edigaryev (approved):

Issue #165, opened 2022-08-01
tart push gets terminated on a large image


victor-yousician:
I set disk size to 200GB
`tart set monterey-base --disk-size 200
`
, extended the disk in the OS
`diskutil apfs resizeContainer disk2 0
`
and pushed the resulting image to ghcr.io and local docker registry. Both tart push commands got terminated with a generic 
`zsh: killed
`
message. From what I see they ran out of memory. I am running this on MacBook Air with only 8GB of memory. Haven't had a chance to see how 32GB memory affects this yet.
Updated: works fine on 32GB. Activity Monitor reports over 50GB memory usage by `tart` process, but it doesn't crash.

fkorotkov:
I think #159 should help with that matter. Let's wait for it to get merged and released.

edigaryev:
Please try out the [new 0.17.0 version](https://github.com/cirruslabs/tart/releases/tag/0.17.0) to see if it fixes the issue for you.

victor-yousician:
I've just tried 0.17.0 - it does not fix the issue, tart still runs out of memory and is killed, but it does bring CPU usage down significantly.

victor-yousician:
Sorry to bug you, but I'm anxiously waiting to try #159. The outcome will decide if I need to request a beefier Mac.

fkorotkov:
@victor-yousician we are still not sure about the root cause. Our guess is that there is a memory leak in the https://github.com/swift-server/async-http-client library that Tart is using for HTTP requests.

We've just release `0.18.0` version with chunked uploads but also with a the latest version of `async-http-client` which by itself might fix your issue. 

If you up for some testing you can try a regular monolithic upload as before with `0.18.0` version which might work just because of the library update.

victor-yousician:
> If you up for some testing you can try a regular monolithic upload as before with `0.18.0` version which might work just because of the library update.

I tried monolithic upload and various chunk sizes (16MB-1024MB) - it makes no difference. Around 96% mark `tart` memory usage balloons and it's terminated.

edigaryev:
> > If you up for some testing you can try a regular monolithic upload as before with `0.18.0` version which might work just because of the library update.
> 
> I tried monolithic upload and various chunk sizes (16MB-1024MB) - it makes no difference. Around 96% mark `tart` memory usage balloons and it's terminated.

[...]

PR #168, opened 2022-08-01
OCI's HTTP client: prevent the usage of NIO Transport Services

edigaryev:
It turns out that:

* the [AsyncHTTPClient](https://github.com/swift-server/async-http-client) package uses [`NIOTSEventLoopGroup()`](https://github.com/swift-server/async-http-client/blob/14fa6d944d88e5702ba630f8a8843966331fd944/Sources/AsyncHTTPClient/HTTPClient.swift#L103) as event loop group
* ...which uses [NIO Transport Services](https://github.com/apple/swift-nio-transport-services)
* ...which in turn uses [`Network.Framework`](https://developer.apple.com/documentation/network)

...which behaves wildly inefficient for reasons that can't be diagnosed easily without getting source code access:

Screenshot 2022-08-02 at 02 08 00

However, we can force the AsyncHTTPClient to take a different path by manually specifying the event loop group provider.

This seems to be a simpler and faster alternative than https://github.com/cirruslabs/tart/pull/159: I'm getting 10x the speed of `main` branch and that of the chunked variant with way less CPU usage.

Resolves https://github.com/cirruslabs/tart/issues/158, https://github.com/cirruslabs/tart/issues/165.

### DIFF
--- a/Sources/tart/OCI/Registry.swift
+++ b/Sources/tart/OCI/Registry.swift
@@ -2,6 +2,7 @@ import Foundation
 import NIOCore
 import NIOHTTP1
 import AsyncHTTPClient
+import NIOPosix
 
 enum RegistryError: Error {
   case UnexpectedHTTPStatusCode(when: String, code: UInt, details: String = "")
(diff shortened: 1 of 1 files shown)


fkorotkov:
Very interesting finding! Thank you for profiling it and finding the root cause. Is there any sense n using more then 1 thread for the  group? Probably not since we don't do parallel requests explicitly but maybe internally there are some events that can be processed in parallel. 🤔

edigaryev:
>Is there any sense n using more then 1 thread for the group? Probably not since we don't do parallel requests explicitly but maybe internally there are some events that can be processed in parallel. 🤔

I don't think it makes sense to bump it at this point since the performance is already quite good and it seems to be a safe choice given that [AsyncHTTPClient uses 1 thread by default](AsyncHTTPClient) too.

fkorotkov (approved):

fkorotkov:
Do we need #159 as a feature now? 😅

PR #169, opened 2022-08-01
OCI: gcr.io fixes

### LINKED ISSUE #166: Unable to use gcr.io as a registry
I'm on tart 0.16.0 and I'm trying to use gcr.io container registry as a registry for VMs. 

I have the credentials json data (same which I use for docker images push/pull so it has necessary access rights) in a file client-secret.json. I'm trying `cat client-secret.json | tart login gcr.io --username _json_key --password-stdin` but it immediately gives me an error `invalid credentials: UnexpectedHTTPStatusCode(when: "doing ping", code: 404, details: "")`

Then I tried with `https://gcr.io` and also with `gcr.io/{project_name}` or `https://gcr.io/{project_name}` - every time it just hangs indefinitely until I hit ^C a few times so it returns `invalid credentials: HTTPClientError.cancelled`.

I believe gcr.io must be OCI-compliant but I'm not 100% sure. 

Am I missing something? Did anyone else had any luck with tart + gcr.io? Any help on this issue would be much appreciated.

edigaryev:
Resolves #166.

### DIFF
--- a/Sources/tart/OCI/Registry.swift
+++ b/Sources/tart/OCI/Registry.swift
@@ -233,7 +233,7 @@ class Registry {
   ) async throws -> HTTPClientResponse {
     var urlComponents = urlComponents
 
-    if urlComponents.queryItems == nil {
+    if urlComponents.queryItems == nil && !parameters.isEmpty {
       urlComponents.queryItems = []
     }
     urlComponents.queryItems?.append(contentsOf: parameters.map { key, value -> URLQueryItem in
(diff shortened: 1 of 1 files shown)


fkorotkov:
Was the 404 caused by the parameters simply being initialized? 🤪

edigaryev:
> Was the 404 caused by the parameters simply being initialized? 🤪

Yeah, it was querying something like `/v2/?`, which caused the 404 😏

fkorotkov (approved):

Issue #131, opened 2022-06-20
CLI integration tests


edigaryev:
So that we can be sure basic command-line combinations like `tart clone`, `tart push` and others work.

fkorotkov:
Let's also include performance testing for pull to catch things like #279

PR #174, opened 2022-08-08
OCI: support insecure registries and custom ports

edigaryev:
Might be helpful for debugging against a local Docker Registry and also for #131.

### DIFF
--- a/Sources/tart/OCI/Registry.swift
+++ b/Sources/tart/OCI/Registry.swift
@@ -104,13 +104,11 @@ class Registry {
   convenience init(
     host: String,
     namespace: String,
+    insecure: Bool = false,
     credentialsProvider: CredentialsProvider = KeychainCredentialsProvider()
   ) throws {
-    var baseURLComponents = URLComponents()
-
-    baseURLComponents.scheme = "https"
-    baseURLComponents.host = host
-    baseURLComponents.path = "/v2/"
+    let proto = insecure ? "http" : "https"
+    let baseURLComponents = URLComponents(string: proto + "://" + host + "/v2/")!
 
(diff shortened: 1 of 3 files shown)


Sources/tart/OCI/Registry.swift
```diff
@@ -104,13 +104,11 @@ class Registry {
...
+    insecure: Bool = false,
     credentialsProvider: CredentialsProvider = KeychainCredentialsProvider()
   ) throws {
-    var baseURLComponents = URLComponents()
-
-    baseURLComponents.scheme = "https"
-    baseURLComponents.host = host
-    baseURLComponents.path = "/v2/"
+    let proto = insecure ? "http" : "https"
+    let baseURLComponents = URLComponents(string: proto + "://" + host + "/v2/")!
```
fkorotkov, line 111: Can we just assume that post might contain either `https` or `http` and do something like:

```swift
var baseURL = host + "/v2/"
if !baseURL.contains("://") {
  baseURL = "https://" + baseURL
}
```

Will it be properly parsed from the upstream invocations?

edigaryev: We'd have to modify the OCI reference parser (`RemoteName.swift`) for this. And an OCI reference parser that accepts URLs and knows anything about protocol seems like a standard non-compliant thing that might bite us in the future.

It also seems weird once we allow `https://ghcr.io/a/b` syntax instead of `ghcr.io/a/b`, it might make someone think they can control the base URL, which this is not the case: the base URL will ignore `/a/b, but will always include `/v2/`.

fkorotkov (approved):

PR #175, opened 2022-08-08
Use Ventura APIs for recovery mode

fkorotkov:
The latest Xcode Beta 5 that just got release removed the private APIs that we used. This change switches to the new APIs plus requires Ventura for running.


.goreleaser.yml
```diff
@@ -32,11 +32,11 @@ brews:
...
-    caveats: See the Github repository for more information
+    caveats: See the GitHub repository for more information
     homepage: https://github.com/cirruslabs/tart
     description: Run macOS VMs on Apple Silicon
     skip_upload: auto
     dependencies:
       - "cirruslabs/cli/softnet"
     custom_block: |
-      depends_on :macos => :monterey
+      depends_on :macos => :ventura
```
edigaryev, line 42: Considering that the Ventura is still in beta until October, don't you think it's too early to drop Monterey support?

We might have multiple of features and bug fixes available in that time frame that will be of use for Monterey users.

[Dynamic](https://github.com/mhdhejazi/Dynamic) might be an option, though.

fkorotkov: But we have a bunch of exclusive features planned on... I was thinking to `ifdef` things around which will work for changes like this but the Linux support and folder sharing will be very hard to maintain IMO.

Another idea might be not to require Ventura but also don't use ifdef. Instead structure code so the codepaths of old features will continue working because of dynamic ducktyping. What do you think?

fkorotkov: See [0e2e74c](https://github.com/cirruslabs/tart/pull/175/commits/0e2e74cbc07a71b3d3457adf69b44f9675ab3d50)

fkorotkov:
@edigaryev if you are still on Monterey. Could you please check the binary after [0e2e74c](https://github.com/cirruslabs/tart/pull/175/commits/0e2e74cbc07a71b3d3457adf69b44f9675ab3d50)

edigaryev:
> @edigaryev if you are still on Monterey. Could you please check the binary after [0e2e74c](https://github.com/cirruslabs/tart/pull/175/commits/0e2e74cbc07a71b3d3457adf69b44f9675ab3d50)

The binary and the recovery option seems to be working 👍

I've Googled a bit about `#available`, and it seems to be [checked at runtime](https://www.avanderlee.com/swift/available-deprecated-renamed/).

This is quite magical because it means that this attribute somehow deals with static linking issue encountered above: once you use some new API it symbol gets added to the binary's symbols table (static linking) and this causes a linking error on older platforms. Dynamic linking is a way to work around with that.

edigaryev (approved):

PR #179, opened 2022-08-12
Do not include Content-Range for monolithic uploads

fkorotkov:
Some registries still assumes it's a chunked upload and verifies the "chunk" size which is too big.

### DIFF
--- a/Sources/tart/OCI/Registry.swift
+++ b/Sources/tart/OCI/Registry.swift
@@ -191,7 +191,7 @@ class Registry {
       )
       if response.status != .created {
         let body = try await response.body.readTextResponse()
-        throw RegistryError.UnexpectedHTTPStatusCode(when: "streaming blob to \(uploadLocation)",
+        throw RegistryError.UnexpectedHTTPStatusCode(when: "pushing blob (PUT) to \(uploadLocation)",
           code: response.status.code, details: body ?? "")
       }
       return digest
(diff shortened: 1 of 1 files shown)


Sources/tart/OCI/Registry.swift
```diff
@@ -177,7 +177,27 @@ class Registry {
...
+        uploadLocation,
+        headers: [
+          "Content-Type": "application/octet-stream",
+        ],
+        parameters: ["digest": digest],
+        body: fromData
+      )
+      if response.status != .created {
+        let body = try await response.body.readTextResponse()
+        throw RegistryError.UnexpectedHTTPStatusCode(when: "streaming blob to \(uploadLocation)",
```
edigaryev, line 194:
```suggestion
        throw RegistryError.UnexpectedHTTPStatusCode(when: "pushing blob (PUT) to \(uploadLocation)",
```

Sources/tart/OCI/Registry.swift
```diff
@@ -177,7 +177,27 @@ class Registry {
...
 
     let digest = Digest.hash(fromData)
+    
+    if chunkSizeMb == 0 {
+      // monolithic upload
+      let response = try await rawRequest(
+        .PUT,
+        uploadLocation,
+        headers: [
+          "Content-Type": "application/octet-stream",
```
edigaryev, line 187:
```suggestion
          "Content-Length": "\(fromData.count)",
          "Content-Type": "application/octet-stream",
```

edigaryev (approved):

edigaryev (approved):

PR #183, opened 2022-08-17
OCI blob compression: fix Data memory leak when using InputFilter

edigaryev:
This makes Tart use `mmap(2)`-backed `Data` and `InputFilter(.compress, ...)` logic from [Apple's documentation](https://developer.apple.com/documentation/accelerate/compressing_and_decompressing_data_with_input_and_output_filters#3328720), which fixes the memory leak and results in smooth memory usage overall:

(this shows a lifetime of a Tart instance that pushed a 120 GB disk to the registry)

Resolves https://github.com/cirruslabs/tart/issues/165.

### DIFF
--- a/Sources/tart/VMDirectory+OCI.swift
+++ b/Sources/tart/VMDirectory+OCI.swift
@@ -125,7 +125,7 @@ extension VMDirectory {
       let data = mappedDisk.subdata(in: mappedDiskReadOffset ..< mappedDiskReadOffset + bytesRead)
       mappedDiskReadOffset += bytesRead
 
-      progress.completedUnitCount += Int64(bytesRead)
+      progress.completedUnitCount = Int64(mappedDiskReadOffset)
 
       return data
     }
--- a/Sources/tart/VMDirectory+OCI.swift
+++ b/Sources/tart/VMDirectory+OCI.swift
@@ -119,11 +119,11 @@ extension VMDirectory {
     // and sequentially upload them as blobs
     let mappedDisk = try Data(contentsOf: diskURL, options: [.alwaysMapped])
     let mappedDiskSize = mappedDisk.count
-    var mappedDiskOffset = 0
+    var mappedDiskReadOffset = 0
     let compressingFilter = try InputFilter(.compress, using: .lz4, bufferCapacity: Self.bufferSizeBytes) { (length: Int) -> Data? in
-      let bytesRead = min(length, mappedDiskSize - mappedDiskOffset)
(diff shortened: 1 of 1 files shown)


Sources/tart/VMDirectory+OCI.swift
```diff
@@ -117,13 +117,15 @@ extension VMDirectory {
...
     // Read VM's compressed disk as chunks
     // and sequentially upload them as blobs
-    let disk = try FileHandle(forReadingFrom: diskURL)
-    var diskReadBytes: UInt64 = 0
-    let compressingFilter = try InputFilter<Data>(.compress, using: .lz4, bufferCapacity: Self.bufferSizeBytes) { _ in
-      let data = try disk.read(upToCount: Self.bufferSizeBytes)
-      diskReadBytes += UInt64(data?.count ?? 0)
+    let mappedDisk = try Data(contentsOf: diskURL, options: [.alwaysMapped])
+    let mappedDiskSize = mappedDisk.count
+    var mappedDiskOffset = 0
```
fkorotkov, line 122: Maybe name it `mappedDiskReadOffset`?

Sources/tart/VMDirectory+OCI.swift
```diff
@@ -117,13 +117,15 @@ extension VMDirectory {
...
+    let compressingFilter = try InputFilter(.compress, using: .lz4, bufferCapacity: Self.bufferSizeBytes) { (length: Int) -> Data? in
+      let bytesRead = min(length, mappedDiskSize - mappedDiskOffset)
+      let data = mappedDisk.subdata(in: mappedDiskOffset ..< mappedDiskOffset + bytesRead)
+      mappedDiskOffset += bytesRead
 
-      progress.completedUnitCount += Int64(data?.count ?? 0)
+      progress.completedUnitCount += Int64(bytesRead)
```
fkorotkov, line 128: Maybe just set it to `mappedDiskOffset`?

Sources/tart/VMDirectory+OCI.swift
edigaryev: See https://github.com/cirruslabs/tart/pull/183/commits/06dc982a9d8416cf1e1543e59ea5f75d3ccf1f55.

Sources/tart/VMDirectory+OCI.swift
edigaryev: See https://github.com/cirruslabs/tart/pull/183/commits/9fdc59729f0ed797a80af275ed338f6c4e713742.

fkorotkov (approved):

PR #199, opened 2022-08-26
Support managing VMs by path

fkorotkov:
Related to https://github.com/hashicorp/vagrant/issues/12760

### DIFF
--- a/Sources/tart/VMStorageLocal.swift
+++ b/Sources/tart/VMStorageLocal.swift
@@ -12,7 +12,12 @@ class VMStorageLocal {
   }
 
   func open(_ name: String) throws -> VMDirectory {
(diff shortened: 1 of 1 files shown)


edigaryev:
What do you think about using [`TART_HOME`](https://github.com/cirruslabs/tart/blob/3a74fc35e676cc70bc70fc3dded61d08a79c0a4a/Sources/tart/Config.swift#L10-L12) to solve this?

fkorotkov:
Good point! We can make the [Vagrant Box Format](https://www.vagrantup.com/docs/plugins/providers) to comply with how `TART_HOME` works (e.g. have `vms` folder inside to be able to do `tart run`).

Let's close the PR for now since we don't have clear picture of the Vagrant integration and seems `TART_HOME` might be used.

PR #202, opened 2022-08-26
Support trackpad on macOS

### LINKED ISSUE #66: handle right click and touch pad gesture in mac books
Hi, thanks for this amazing tools, i am using `tart` to test some isolated build of c++ programs locally, and found that i wasn't able to send a right click using the usual two-finger click in the vm.

fkorotkov:
Fixes #66

### DIFF
--- a/Sources/tart/Platform/Platform.swift
+++ b/Sources/tart/Platform/Platform.swift
@@ -5,4 +5,5 @@ protocol Platform: Codable {
     func bootLoader(nvramURL: URL) throws -> VZBootLoader
     func platform(nvramURL: URL) -> VZPlatformConfiguration
     func graphicsDevice(vmConfig: VMConfig) -> VZGraphicsDeviceConfiguration
+    func pointingDevices(vmConfig: VMConfig) -> [VZPointingDeviceConfiguration]
 }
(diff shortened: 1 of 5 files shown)


Sources/tart/Platform/Platform.swift
```diff
@@ -5,4 +5,5 @@ protocol Platform: Codable {
     func bootLoader(nvramURL: URL) throws -> VZBootLoader
     func platform(nvramURL: URL) -> VZPlatformConfiguration
     func graphicsDevice(vmConfig: VMConfig) -> VZGraphicsDeviceConfiguration
+    func pointingDevices(vmConfig: VMConfig) -> [VZPointingDeviceConfiguration]
```
edigaryev, line 8: You can probably can skip the `vmConfig` argument here as it's unused.

fkorotkov: Good catch! See [c1d7ebb](https://github.com/cirruslabs/tart/pull/202/commits/c1d7ebb5ffaa52230d5a37c393e51cbe43d3ee18).

edigaryev:
The code LGTM, however, I'm not sure if features like clipboard sharing and similar should be enabled by default due to security concerns.

For example, in the cloud two VMs would be able to manipulate each other's clipboard contents (this currently applies to Linux, but I can imagine Apple adding the very same feature for macOS at some point) and read host's clipboard contents.

I can also see some users not wanting this enabled by default: those who also run multiple VMs on a host and offering some sort of service (multi-tenant) or those who are simply wishing for more isolation on their own machine (single-tenant).

What do you think about adding a `tart run --clipboard` flag to enable this?

fkorotkov:
Good point. Let's postpone until macOS also supports clipboard sharing. Reverted.

edigaryev (approved):

PR #205, opened 2022-08-29
Support Docker Helpers

### LINKED ISSUE #167: Support credentials helpers
Will be nice to support [credentials helpers](https://docs.docker.com/engine/reference/commandline/login/#credential-helper-protocol). Can make Tart a bit smarter and check `$HOME/.docker/config.json` for the config`

Related issues: #166

fkorotkov:
Fixes #167

### DIFF
--- a/Sources/tart/Credentials/HelperProgramCredentialsProvider.swift
+++ b/Sources/tart/Credentials/HelperProgramCredentialsProvider.swift
@@ -1,0 +1,63 @@
+import Foundation
+
+class HelperProgramCredentialsProvider: CredentialsProvider {
+  func retrieve(host: String) throws -> (String, String)? {
+    let dockerConfigURL = FileManager.default.homeDirectoryForCurrentUser.appendingPathComponent(".docker").appendingPathComponent("config.json")
+    if !FileManager.default.fileExists(atPath: dockerConfigURL.path) {
+      return nil
+    }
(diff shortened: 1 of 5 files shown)


Sources/tart/Credentials/HelperProgramCredentialsProvider.swift
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    let getOutput = try JSONDecoder().decode(
+      DockerGetOutput.self, from: outPipe.fileHandleForReading.readDataToEndOfFile()
+    )
+    return (getOutput.Username, getOutput.Secret)
+  }
+
+  func store(host: String, user: String, password: String) throws {
+    throw CredentialsProviderError.Failed(message: "Docker helpers don't support storing!")
```
edigaryev, line 52: I think we need to fallback to `KeychainCredentialsProvider()` here, otherwise this effectively prevents the `tart login` from working on computers that have Docker configured to use credential helpers.

fkorotkov: `login` is already using only keychain. See https://github.com/cirruslabs/tart/blob/048a5506df871aaddc412ddfabe2e52bec850e57/Sources/tart/Commands/Login.swift#L56

edigaryev (approved):

Issue #13, opened 2022-03-28
Support mounting of shared folders


fkorotkov:
Seems it's doable. See https://github.com/gyf304/vmcli/issues/20#issuecomment-867349378

fkorotkov:
@edigaryev it seems once macOS Ventura is in public beta we can try to support it. There is new [macOSGuestAutomountTag](https://developer.apple.com/documentation/virtualization/vzvirtiofilesystemdeviceconfiguration/3969279-macosguestautomounttag) which seems indicates folder mounting will work.

PR #211, opened 2022-08-29
Directory sharing support

edigaryev:
Works both on macOS and Linux.

Requires macOS Ventura.

Resolves https://github.com/cirruslabs/tart/issues/13.

### DIFF
--- a/Sources/tart/Commands/Run.swift
+++ b/Sources/tart/Commands/Run.swift
@@ -37,12 +37,20 @@ struct Run: AsyncParsableCommand {
   @Flag var withSoftnet: Bool = false
 
   @Option(help: ArgumentHelp("""
-    Additional disk attachments with an optional read-only specifier\n(e.g. --disk=\"disk.bin\" --disk=\"disk.bin:ro\")
+    Additional disk attachments with an optional read-only specifier\n(e.g. --disk=\"disk.bin\" --disk=\"ubuntu.iso:ro\")
     """, discussion: """
     Learn how to create a disk image using Disk Utility here:
     https://support.apple.com/en-gb/guide/disk-utility/dskutl11888/mac
-    """))
+    """, valueName: "path[:ro]"))
   var disk: [String] = []
+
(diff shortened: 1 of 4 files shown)


Sources/tart/Commands/Run.swift
```diff
@@ -37,13 +37,21 @@ struct Run: AsyncParsableCommand {
...
-    """))
+    """, valueName: "path[:ro]"))
   var disk: [String] = []
 
+  @Option(help: ArgumentHelp("""
+                             Additional directory shares with an optional read-only specifier\n(e.g. --dir=\"build:~/src/build\" --dir=\"sources:~/src/sources:ro\")
+                             """, discussion: """
+                                              All shared directories are automatically mounted to "/Volumes/My Shared Files" directory on macOS,
+                                              while on Linux you have to do it manually: "mount -t virtiofs com.apple.virtio-fs.automount /mount/point".
+                                              """, valueName: "name:path[:ro]"))
```
fkorotkov, line 52: Why there is a need for a name?

edigaryev: [`VZMultipleDirectoryShare`](https://developer.apple.com/documentation/virtualization/vzmultipledirectoryshare) requires names to guarantee collision-free mounting of multiple directories to a single mountpoint.

Sources/tart/VM.swift
```diff
@@ -278,6 +282,18 @@ class VM: NSObject, VZVirtualMachineDelegate, ObservableObject {
     // Entropy
     configuration.entropyDevices = [VZVirtioEntropyDeviceConfiguration()]
 
+    // Directory share
+    if #available(macOS 13, *) {
```
fkorotkov, line 286: What do you think throwing an error on Monterey?

edigaryev: See https://github.com/cirruslabs/tart/pull/211/commits/01884f7021edb87b6a193c973704dd60950c7c95.

fkorotkov (approved):

PR #215, opened 2022-08-30
tart clone: clone VM and generate MAC under a file lock

edigaryev:
Otherwise there exists a race condition that might result in two parallel `tart clone` instances creating two VMs with a similar MAC addresses.

### DIFF
--- a/Sources/tart/Commands/Clone.swift
+++ b/Sources/tart/Commands/Clone.swift
@@ -32,12 +32,17 @@ struct Clone: AsyncParsableCommand {
       }
 
       let sourceVM = try VMStorageHelper.open(sourceName)
-      let generateMAC = try localStorage.hasVMsWithMACAddress(macAddress: sourceVM.macAddress())
 
       let tmpVMDir = try VMDirectory.temporary()
       try await withTaskCancellationHandler(operation: {
+        let lock = try FileLock(lockURL: Config().tartHomeDir)
+        try lock.lock()
+
+        let generateMAC = try localStorage.hasVMsWithMACAddress(macAddress: sourceVM.macAddress())
         try sourceVM.clone(to: tmpVMDir, generateMAC: generateMAC)
         try localStorage.move(newName, from: tmpVMDir)
+
+        try lock.unlock()
       }, onCancel: {
         try? FileManager.default.removeItem(at: tmpVMDir.baseURL)
       })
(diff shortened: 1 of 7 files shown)


fkorotkov (commented):
I wonder if that's the right place to put the lock. Especially a file lock... What if something happens and `tart clone` invocation will fail to remove a lock... Then the host will become unusable and all sequential clones will just lock.

IMO looks like it's more responsibility of the called to take care of the locking. Maybe just add a simple in-memory lock via a mutex to [`NewVMClonedFrom`](https://github.com/cirruslabs/cirrus-cli/blob/5153b7e532c9b643d1a6b2bcf2573ae6308400f5/internal/executor/instance/persistentworker/isolation/tart/vm.go#L23)? That way there will be no issue with two parallel `pull`s as well. 🤔

edigaryev:
> What if something happens and `tart clone` invocation will fail to remove a lock... Then the host will become unusable and all sequential clones will just lock.

[`flock(2)`](https://man.openbsd.org/flock) lock is released once the process that holds it terminates.

> IMO looks like it's more responsibility of the called to take care of the locking. Maybe just add a simple in-memory lock via a mutex to [`NewVMClonedFrom`](https://github.com/cirruslabs/cirrus-cli/blob/5153b7e532c9b643d1a6b2bcf2573ae6308400f5/internal/executor/instance/persistentworker/isolation/tart/vm.go#L23)? That way there will be no issue with two parallel `pull`s as well. 🤔

Locking this from CLI is an option, yes, however, I'm not sure if that'd be appropriate to fix the Tart's specific issue in Cirrus CLI, considering the `flock(2)`'s property I've mentioned above.

fkorotkov:
What about the concern about two parallel pulls from the CLI side?

edigaryev:
> What about the concern about two parallel pulls from the CLI side?

[`tart pull` operations are atomic](https://github.com/cirruslabs/tart/pull/109) since relatively recently, or do you have some other concerns in mind?

fkorotkov:
What will happen if a persistent worker will get two tasks that will require to pull a new hefty images? Seems with the current logic two tart instances will actually pull the image twice in parallel to only fail once it will try to rename. 🤔

edigaryev:
> What will happen if a persistent worker will get two tasks that will require to pull a new hefty images? Seems with the current logic two tart instances will actually pull the image twice in parallel to only fail once it will try to rename. 🤔

Added a lock on concurrent `tart pull`'s for the same registry host in https://github.com/cirruslabs/tart/pull/215/commits/1b985a22697094ac6534e667092ae508f1a30fb9 as discussed, PTAL.

fkorotkov (approved):

PR #223, opened 2022-09-01
Fix cache reclaimed bytes calculation

edigaryev:
To prevent accidental deletion of all cache entries 🤦‍♂️


fkorotkov (approved):

fkorotkov:
FYI @remarkov and @mikhail-tokarev in case your integration is also trying to pull every time. There is a logic where if there is not enough space on host, Tart is trying to prune older versions of VMs. This logic had a bug and this PR fixes it.

PR #225, opened 2022-09-05
Improve RemoteName parser

### LINKED ISSUE #35: Improve RemoteName parser
OCI image names are widely used, but a formal spec on their format is missing. Here's what what [OCI's Unique Resource Identifiers](https://github.com/opencontainers/.github/blob/master/docs/docs/introduction/digests.md#unique-resource-identifiers) says on the subject, for example:

>How do we ask for containers? We ask for a unique resource identifier (uri) which broadly comes in this form:

There's nothing formal in the entire document, except that the resource identifier is split into `registry`, `organization` and `repository`.

There's an implementation in [Docker's distribution package](https://github.com/distribution/distribution/blob/main/reference/regexp.go), which is way more complicated than our's `RemoteName` parser.

There's [an attempt to derive an EBNF grammar](https://ktomk.github.io/pipelines/doc/DOCKER-NAME-TAG.html#grammar) which is quite readable at first glance (compared to a bunch of regular expression above), so perhaps we should validate it and stick to it.

edigaryev:
Resolves #35.


Sources/tart/OCI/Reference/Reference.g4
```diff
@@ -0,0 +1,13 @@
+grammar Reference;
```
fkorotkov, line 1: What the origin of this grammar? You compiled it from a spec? Would be nice to have a link here.

edigaryev: No, it's a port of the old grammar that Tart used, but more (1) readable/editable and (2) strict w.r.t. potential path traversal (see d8a71ec).

fkorotkov (approved):

PR #228, opened 2022-09-06
Clarify if guest or host is unsupported

raven:
Whilst attempting to use the directory share support, I encountered:  
`Error: error: directory sharing is only supported on macOS 13.0 (Ventura) or newer`

From that message it wasn't abundantly clear if it was my host or my guest that was not running the supported version of macOS

Issue #59, opened 2022-05-10
Custom Serial Injection


ChefAustin:
There are some times when it is beneficial to be able to control the VM's presented serial number. 

This is a feature request to allow for custom serial injection on a Tart VM.

fkorotkov:
I did some investigation and it seems the only places we can configure something hardware related is via binary representation of `VZMacMachineIdentifier` stored in `ecid` field of `config.json` and `VZMacHardwareModel` stored in `hardwareModel` field respectively. There are no public APIs so we can only try to modify the binary representation at the moment. 😅

Data stored there is stored in Apple's binary plist format and can be decoded:

```shell
pbpaste | base64 --decode > ecid.plist
plutil -convert xml1 ecid.plist

pbpaste | base64 --decode > hardwareModel.plist
plutil -convert xml1 hardwareModel.plist
```

Here are the contents of the files for our [`monterey-base`](https://github.com/orgs/cirruslabs/packages/container/package/macos-monterey-base) VM:

```xml
<plist version="1.0">
<dict>
	<key>ECID</key>
	<integer>10796302860778061261</integer>
</dict>
</plist>
```

```xml
<plist version="1.0">
<dict>
	<key>DataRepresentationVersion</key>
	<integer>1</integer>
	<key>MinimumSupportedOS</key>
	<array>
		<integer>12</integer>
		<integer>0</integer>
		<integer>0</integer>
	</array>
	<key>PlatformVersion</key>
	<integer>2</integer>
</dict>
</plist>
```

It seems at the moment it's not quite possible to configure it. Maybe it's possible with some secret key which we don't know yet.

WWDC is just 4 weeks away and hopefully there will be more news about `Virtualization.Framework` so let's wait until then and do more investigation after WWDC.

TheEagle13:
I understand there is currently not the possibility of setting a specific Serial Number, but is there any method to randomly generate a new one? I'm attempting to setup a workflow which creates VM Templates, but for obvious reasons, There needs to be a new serial on the VMs deployed from the template.

edigaryev:
> I understand there is currently not the possibility of setting a specific Serial Number, but is there any method to randomly generate a new one? I'm attempting to setup a workflow which creates VM Templates, but for obvious reasons, There needs to be a new serial on the VMs deployed from the template.

[...]

PR #231, opened 2022-09-07
Clarify requirements for `dir` argument

raven:
Follow on from https://github.com/cirruslabs/tart/pull/228

It is unclear when using the `dir` directory sharing that it requires guests to be running macOS 13 also. This PR tries to clarify that requirement in the help discussion output. 

Is there a way to read back out from the VM either running or prior to booting to know the version macOS? Would be nicer to report the failure as opposed to requiring the user read the docs.

### DIFF
--- a/Sources/tart/Commands/Run.swift
+++ b/Sources/tart/Commands/Run.swift
@@ -47,8 +47,10 @@ struct Run: AsyncParsableCommand {
   @Option(help: ArgumentHelp("""
                              Additional directory shares with an optional read-only specifier\n(e.g. --dir=\"build:~/src/build\" --dir=\"sources:~/src/sources:ro\")
                              """, discussion: """
+                                              Requires host to be macOS 13.0 (Ventura) or newer.
                                               All shared directories are automatically mounted to "/Volumes/My Shared Files" directory on macOS,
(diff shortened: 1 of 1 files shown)


edigaryev (approved):

edigaryev:
> Is there a way to read back out from the VM either running or prior to booting to know the version macOS? Would be nicer to report the failure as opposed to requiring the user read the docs.

While it would be possible to write the [macOS version](https://developer.apple.com/documentation/virtualization/vzmacosrestoreimage/3816078-operatingsystemversion) at the VM installation step to the VM's configuration file and then consult it to figure out whether the `--dir` is supported.

However, it is still unclear what to do with VM's that were created without this logic. We can simply throw an error in this case and require the user to manually specify macOS version using `tart set`. Afterwards the `--dir` will continue to work.

@fkorotkov what do you think?

fkorotkov (approved):

fkorotkov:
I just checked and [`hardwareModel`](https://github.com/cirruslabs/tart/issues/59#issuecomment-1123711678) for Ventura VMs has min OS version set to 12 which is expected since you can run Ventura guests on Monterey hosts. So I guess there is no way to easily figure out the macOS version by just looking at an image.

@edigaryev suggestion might not work since one can initialize an OS update from within a VM. 🤔

Issue #232, opened 2022-09-07
VM fails to start if `dir` is given a path with `~`


raven:
## Summary

Whilst using the new `dir` option for directory shares, it seems that VM fails to start if `dir` is given a path with `~`, e.g `tart run --dir="downloads:~/Downloads" base`

| Invocation | Result |
| --- | --- |
| `tart run --dir="downloads:~/Downloads" base` | ❌ Hangs |
| `tart run --dir="downloads:Downloads" base` | ✅ Launches |
| `tart run --dir="downloads:/Users/peter.goldsmith/Downloads" base` | ✅ Launches |

I think I can see where needs to be updated. `URL(fileURLWithPath: path)` doesn't automatically handle and map tilde. 

### Environment

Host: macOS 13 Ventura beta 6
Guest: macOS 13 Ventura beta 6

PR #233, opened 2022-09-07
Handle tilde in path for directory share

raven:
See https://github.com/cirruslabs/tart/issues/232

If you create a `URL` such as  `URL(fileURLWithPath: "~/Downloads")` it doesn't automatically expand the tilde into the home directory. 

This PR uses NSString's `expandingTildeInPath` api to do the heavy lifting. 

Ideally `VZVirtualMachine` wouldn't hang and instead log if the directory share paths given didn't exist.

### DIFF
--- a/Sources/tart/Commands/Run.swift
+++ b/Sources/tart/Commands/Run.swift
@@ -166,7 +166,7 @@ struct Run: AsyncParsableCommand {
 
       let (name, path) = (String(splits[0]), String(splits[1]))
 
-      result.append(DirectoryShare(name: name, path: URL(fileURLWithPath: path), readOnly: readOnly))
+      result.append(DirectoryShare(name: name, path: URL(fileURLWithPath: NSString(string: path).expandingTildeInPath), readOnly: readOnly))
     }
 
     return result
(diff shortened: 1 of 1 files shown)


raven:
I'm happy to add validation ensuring that provided paths exist. Is there any use case (Linux maybe) where a user would need to reference a share directory that doesn't exist on the host?

edigaryev (approved):
LGTM, would be nice to add tilde expansion in the other places too:

* `tart create --from-ipsw <path>`
* `tart run --disk <path[:ro]>`

But we can do this in a separate PR.

raven:
> tart create --from-ipsw <path>

Interestingly tilde is no issue here. The [consuming API](https://github.com/cirruslabs/tart/blob/main/Sources/tart/VM.swift#L125) of [`VZMacOSRestoreImage.load(from:completionHandler:)](https://developer.apple.com/documentation/virtualization/vzmacosrestoreimage/3816620-load) seems to handle URL's constructed with `~` just fine. I'm guessing they differ internally with how they're handling URLs. 

> tart run --disk <path>

I'll update this one. Interestingly, again there is different behaviour. If fed a URL with a tilde, which will map to an absoluteString that doesn't exist, `VZDiskImageStorageDeviceAttachment(url:readOnly:)` actually throws and you're given the error: `"No such file or directory"`, which bubbles up to the tart user in the form: `Error: The operation couldn’t be completed. No such file or directory`

edigaryev (approved):

fkorotkov (approved):

fkorotkov:
Thank you! ❤️

Issue #222, opened 2022-09-01
No space left on device


magnusviri:
I just made an issue regarding moving the images to an external drive. I guess tart is using the system cache, which is on the drive still. It filled my hard drive (250 GB).

```
Error Domain=NSCocoaErrorDomain Code=640 "“disk.img” couldn’t be copied to “DA30BDDE-FD5B-4469-ADD8-9E5454BDFEC8” because there isn’t enough space." UserInfo={NSSourceFilePathErrorKey=/Volumes/autopkg/tart/vms/monterey-base/disk.img, NSUserStringVariant=(
    Copy
), NSDestinationFilePath=/var/folders/m9/pqzn80q548z2419lljbgydh40000gp/T/DA30BDDE-FD5B-4469-ADD8-9E5454BDFEC8/disk.img, NSFilePath=/Volumes/autopkg/tart/vms/monterey-base/disk.img, NSUnderlyingError=0x60000236ed90 {Error Domain=NSPOSIXErrorDomain Code=28 "No space left on device"}}
```

edigaryev:
It seems that this is due to temporary directory's usage, which is used to guarantee atomicity for some operations.

Can you list some steps to reproduce your situation? I.e. which tart commands are you running (I assume `tart clone`?) and on which images (mostly interested in their sizes on disk).

magnusviri:
This is a condensed history.

```
brew install cirruslabs/cli/tart
tart clone ghcr.io/cirruslabs/macos-monterey-base:latest monterey-base
```

Here I notice I don't have enough hard drive space

```
mv .tart /Volumes/autopkg/tart
```

I tried symlinking and cloning again but it failed. I set the env var

```
tart clone ghcr.io/cirruslabs/macos-monterey-base:latest autopkg
tart list
```

And this is my VM list

```
local	monterey-base
oci	ghcr.io/cirruslabs/macos-monterey-base@sha256:309b7466f23049e4594c4e9a6370e4e5906796a9a3c210ce1b8ad448d8da27ff
```

Here's the condensed output of ls -lRh.

```
/Volumes/autopkg/tart/cache/OCIs/ghcr.io/cirruslabs/macos-monterey-base/sha256:309b7466f23049e4594c4e9a6370e4e5906796a9a3c210ce1b8ad448d8da27ff:
total 58659328
-rw-r--r--  1 james  staff    28G Sep  1 10:28 disk.img

/Volumes/autopkg/tart/vms/monterey-base:
total 58659328
[...]

PR #238, opened 2022-09-11
Self-hosted temporary directory

edigaryev:
Resolves https://github.com/cirruslabs/tart/issues/222.


fkorotkov:
Should `VMDirectory.temporary` just acquire the lock?

edigaryev:
> Should `VMDirectory.temporary` just acquire the lock?

No, it will release it then. I've tried to wrap it before but it turned out to be more complicated than it should be as you need to dereference the original `VMDirectory` object from the wrapper.

fkorotkov (approved):

Issue #227, opened 2022-09-05
Feature request: `tart set <name> --name <newname>`


magnusviri:
I'd like to change the name of my VM.

PR #246, opened 2022-09-14
Introduce "tart rename" command to rename VMs

edigaryev:
Resolves https://github.com/cirruslabs/tart/issues/227.

### DIFF
--- a/Sources/tart/Commands/Rename.swift
+++ b/Sources/tart/Commands/Rename.swift
@@ -1,0 +1,41 @@
+import ArgumentParser
+import Foundation
+import SystemConfiguration
+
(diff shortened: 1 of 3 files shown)


Sources/tart/Commands/Rename.swift
```diff
@@ -0,0 +1,40 @@
+import ArgumentParser
+import Foundation
+
+struct Rename: AsyncParsableCommand {
+    static var configuration = CommandConfiguration(abstract: "Clone a VM")
```
raven, line 5:
```suggestion
    static var configuration = CommandConfiguration(abstract: "Rename a VM")
```

fkorotkov (approved):

Issue #243, opened 2022-09-13
Bridged Network Support


fkorotkov:
We got approved for the `com.apple.vm.networking` entitlement so now we can use [`VZBridgedNetworkDeviceAttachment`](https://developer.apple.com/documentation/virtualization/vzbridgednetworkdeviceattachment).

FYI @naikrovek @magnusviri

naikrovek:
oh, wow, nice

cosmoshepcat:
now if apple would just support snapshots in their virtual framework :)

fkorotkov:
@cosmoshepcat you mean snapshot like in #147?

cosmoshepcat:
> @cosmoshepcat you mean snapshot like in #147?

sort of, what i meant is apple does not have any such core vm snapshot feature in the virtualization framework that I know of.  this is why other commercial products like parallels etc cannot take vm snapshots like they do on intel versions/previous versions of their software since they now rely on the api framework for silicon m1 support.

i do not know apples roadmap for such features if any

fkorotkov:
We had to revert the entitlements in #247 since they require an embedded provisioning profile and we haven't yet figure out how to do it for a CLI toll and not an `*.App`.

fkorotkov:
From this [doc](https://developer.apple.com/documentation/technotes/tn3125-inside-code-signing-provisioning-profiles):

> Some macOS products, like daemons and command-line tools, ship as a standalone executable. A standalone executable can’t claim a restricted entitlement because there’s no place to embed the provisioning profile that authorizes that claim. If your standalone executable needs to do this, wrap it in an app-like structure. For an example of this, see [Signing a Daemon with a Restricted Entitlement](https://developer.apple.com/documentation/Xcode/signing-a-daemon-with-a-restricted-entitlement).

`If your standalone executable needs to do this, wrap it in an app-like structure.` part is very concerning since Tart is a single binary at the moment. Hopefully there is another way...

fkorotkov:
It seems the only way right now to embed a provisioning profile is to distribute Tart as an `*.App` and install it in `/Applications`. 😪🤷‍♂️

cosmoshepcat:
apple...  :(

edigaryev:
> If your standalone executable needs to do this, wrap it in an app-like structure

What if we would ship an App as an option, though?

Seems like it would be possible to handle this via Homebrew. Here's an example of `vlc` installed via `brew`:

```
[...]

PR #247, opened 2022-09-14
Revert com.apple.vm.networking entitlement

edigaryev:
While we figure out what's wrong with the `com.apple.vm.networking`.

### DIFF
--- a/Resources/tart.entitlements
+++ b/Resources/tart.entitlements
@@ -4,7 +4,5 @@
 <dict>
 	<key>com.apple.security.virtualization</key>
 	<true/>
-	<key>com.apple.vm.networking</key>
-	<true/>
 </dict>
 </plist>


fkorotkov (approved):

inez:
Quick question: it seems that without this entitlement the bridge modes does not work - do you mind clarifying how/in what way it was broken?
We are looking for a way to share network (and VPN connection) with the host machine, and seems like that (bridged mode) might be the way.

edigaryev:
> Quick question: it seems that without this entitlement the bridge modes does not work - do you mind clarifying how/in what way it was broken?

See https://github.com/cirruslabs/tart/issues/243#issuecomment-1248112324.

> We are looking for a way to share network (and VPN connection) with the host machine, and seems like that (bridged mode) might be the way.

You can already do this using the default NAT networking. The packets from the guest VM will in the end use the default gateway on the host, meaning that they'll be routed via the VPN, if configured properly.

PR #248, opened 2022-09-19
Set line-buffered output for stdout and introduce --graphics for tart run

edigaryev:
Both of these changes are required for https://github.com/cirruslabs/tart/issues/101.

The line-buffering enables us to read the VNC-server address from Packer in-time, otherwise we might not see the VNC-line because to macOS/Swift using fully-buffered streams. I've placed the [`setlinebuf(3)`](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/setlinebuf.3.html) in `Root.swift` because we might eventually stumble upon the same issue with some other Tart sub-command.

The other change enables us to use `headless` configuration option in Packer to debug the VNC automation by looking at what it does.

### DIFF
--- a/Sources/tart/Commands/Run.swift
+++ b/Sources/tart/Commands/Run.swift
@@ -15,7 +15,7 @@ struct Run: AsyncParsableCommand {
   var name: String
 
   @Flag(help: ArgumentHelp(
-          "Don't open an UI window.",
+          "Don't open a UI window.",
           discussion: "Useful for integrating Tart VMs into other tools.\nUse `tart ip` in order to get an IP for SSHing or VNCing into the VM.")) 
   var noGraphics: Bool = false
 


Sources/tart/Commands/Run.swift
```diff
@@ -15,10 +15,13 @@ struct Run: AsyncParsableCommand {
   var name: String
 
   @Flag(help: ArgumentHelp(
-          "Don't open a UI window.",
+          "Don't open an UI window.",
```
raven, line 18:
```suggestion
          "Don't open a UI window.",
```
Only need to use 'an' before a vowel sound, not a vowel letter https://www.thesaurus.com/e/grammar/a-vs-an/

Sources/tart/Commands/Run.swift
```diff
@@ -15,10 +15,13 @@ struct Run: AsyncParsableCommand {
...
           discussion: "Useful for integrating Tart VMs into other tools.\nUse `tart ip` in order to get an IP for SSHing or VNCing into the VM.")) 
   var noGraphics: Bool = false
 
+  @Flag(help: "Force open an UI window, even when VNC is enabled.")
```
raven, line 22:
```suggestion
  @Flag(help: "Force open a UI window, even when VNC is enabled.")
```
Same as other comment.

Issue #239, opened 2022-09-12
Log important events


edigaryev:
[Like this one](https://github.com/cirruslabs/tart/blob/f38b1aed9d7436dbc0d5721f5eed344d9f7a64aa/Sources/tart/VMStorageOCI.swift#L157-L167), to `$TART_HOME/tart.log`, for example.

Logging libraries:

* https://github.com/apple/swift-log
* https://github.com/sushichop/Puppy

fkorotkov:
Will help to deal with suspicious like #240. We should log things like why a prunning or gc process started.

PR #265, opened 2022-10-07
Log cache pruning to $TART_HOME/tart.log

edigaryev:
Resolves https://github.com/cirruslabs/tart/issues/239.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -10,6 +10,15 @@
       }
     },
     {
+      "identity" : "puppy",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/sushichop/Puppy",
+      "state" : {
+        "revision" : "3e8d87f714f14244878752a6bb71ea465119f8a1",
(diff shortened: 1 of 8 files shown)


fkorotkov (commented):
Puppy seems to be able to act as a [backend for SwiftLog](https://github.com/sushichop/Puppy#use-with-appleswift-log). I think `SwiftLog` should be used everyever instead of `Puppy.default`, isnt's it?

edigaryev:
> Puppy seems to be able to act as a [backend for SwiftLog](https://github.com/sushichop/Puppy#use-with-appleswift-log). I think `SwiftLog` should be used everyever instead of `Puppy.default`, isnt's it?

It's not strictly required:

>It not only works alone, but also as a backend for [apple/swift-log](https://github.com/apple/swift-log/).

This is from [Puppy's `README.md`](https://github.com/sushichop/Puppy#use-with-appleswift-log).

fkorotkov (approved):

Issue #271, opened 2022-10-13
`tart prune --cache-budget` pruned my whole cache


adrienkohlbecker:
Hi! 

It seems that the `--cache-budget` argument in `tart prune` is not doing what it is supposed to do. In the example below, it deleted my whole cache:

```
# state of cache before
ec2-user@ip-172-31-12-38 ~ % du -h -d1 .tart/cache/OCIs/ACCOUNT_ID.dkr.ecr.eu-west-1.amazonaws.com/macos
 84G    .tart/cache/OCIs/ACCOUNT_ID.dkr.ecr.eu-west-1.amazonaws.com/macos/repo-1
 84G    .tart/cache/OCIs/ACCOUNT_ID.dkr.ecr.eu-west-1.amazonaws.com/macos/repo-2
 47G    .tart/cache/OCIs/ACCOUNT_ID.dkr.ecr.eu-west-1.amazonaws.com/macos/repo-3
168G    .tart/cache/OCIs/ACCOUNT_ID.dkr.ecr.eu-west-1.amazonaws.com/macos/repo-4
382G    .tart/cache/OCIs/ACCOUNT_ID.dkr.ecr.eu-west-1.amazonaws.com/macos

# command executed
ec2-user@ip-172-31-12-38 ~ % tart prune --cache-budget=350

# state of cache after 
ec2-user@ip-172-31-12-38 ~ % du -h -d1 .tart/cache/OCIs/ACCOUNT_ID.dkr.ecr.eu-west-1.amazonaws.com/macos
  0B    .tart/cache/OCIs/ACCOUNT_ID.dkr.ecr.eu-west-1.amazonaws.com/macos/repo-1
  0B    .tart/cache/OCIs/ACCOUNT_ID.dkr.ecr.eu-west-1.amazonaws.com/macos/repo-2
  0B    .tart/cache/OCIs/ACCOUNT_ID.dkr.ecr.eu-west-1.amazonaws.com/macos/repo-3
  0B    .tart/cache/OCIs/ACCOUNT_ID.dkr.ecr.eu-west-1.amazonaws.com/macos/repo-4
  0B    .tart/cache/OCIs/ACCOUNT_ID.dkr.ecr.eu-west-1.amazonaws.com/macos

# version
ec2-user@ip-172-31-12-38 ~ % tart --version
0.29.0
```

adrienkohlbecker:
Thanks! Much appreciated. I will test this with the future release

PR #272, opened 2022-10-13
Fix tart prune --cache-budget logic

edigaryev:
We now try to gather as much of the recent cache entries as possible that fit the budget before deleting the rest.

Resolves https://github.com/cirruslabs/tart/issues/271.

### DIFF
--- a/Sources/tart/Commands/Prune.swift
+++ b/Sources/tart/Commands/Prune.swift
@@ -63,21 +63,22 @@ struct Prune: AsyncParsableCommand {
     let prunableStorages: [PrunableStorage] = [VMStorageOCI(), try IPSWCache()]
     let prunables: [Prunable] = try prunableStorages
             .flatMap { try $0.prunables() }
-            .sorted { try $0.accessDate() < $1.accessDate() }
+            .sorted { try $0.accessDate() > $1.accessDate() }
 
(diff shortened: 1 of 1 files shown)


Sources/tart/Commands/Prune.swift
```diff
@@ -63,21 +63,22 @@ struct Prune: AsyncParsableCommand {
...
 
-    var it = prunables.makeIterator()
+    for prunable in prunables {
+      let prunableSizeBytes = UInt64(try prunable.sizeBytes())
 
-    while (cacheUsedBytes - cacheReclaimedBytes) > cacheBudgetBytes {
-      guard let prunable = it.next() else {
-        break
+      if prunableSizeBytes <= cacheBudgetBytes {
+        cacheBudgetBytes -= prunableSizeBytes
```
fkorotkov, line 75: Let's add a comment so it's easier to understand. Maybe something like:

```suggestion
        // don't mark for deletion while there is budged
        cacheBudgetBytes -= prunableSizeBytes
```

edigaryev: See https://github.com/cirruslabs/tart/pull/272/commits/011369d2148d8d0d0d2f00cdb7a9f0d2d6fc7116.

fkorotkov (approved):

PR #274, opened 2022-10-14
tart run: deprecate --with-softnet and introduce --net-softnet

edigaryev:
For consistency's sake with `--net-bridged`.

### DIFF
--- a/Sources/tart/Commands/Run.swift
+++ b/Sources/tart/Commands/Run.swift
@@ -37,7 +37,8 @@ struct Run: AsyncParsableCommand {
       + "Note that this feature is experimental and there may be bugs present when using VNC."))
   var vncExperimental: Bool = false
 
-  @Flag var withSoftnet: Bool = false
+  @Flag(help: ArgumentHelp(visibility: .private))
+  var withSoftnet: Bool = false
 
   @Option(help: ArgumentHelp("""
     Additional disk attachments with an optional read-only specifier\n(e.g. --disk=\"disk.bin\" --disk=\"ubuntu.iso:ro\")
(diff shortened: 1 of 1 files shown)


Sources/tart/Commands/Run.swift
```diff
@@ -64,6 +65,10 @@ struct Run: AsyncParsableCommand {
                                               """, valueName: "interface name"))
   var netBridged: String?
 
+  @Flag(help: ArgumentHelp("Use software networking instead of the default shared (NAT) networking",
+          discussion: "Learn how to configure Softnet for use with Tart here: https://github.com/cirruslabs/softnet"))
+  var netSoftnet: Bool = false
+
   func validate() throws {
```
fkorotkov, line 72: Maybe just make `netSoftnet = netSoftnet  || withSoftnet` to respect the old value and then change all occurrences of `withSoftnet` to `netSoftnet`?

Sources/tart/Commands/Run.swift
```diff
@@ -37,7 +37,8 @@ struct Run: AsyncParsableCommand {
       + "Note that this feature is experimental and there may be bugs present when using VNC."))
   var vncExperimental: Bool = false
 
-  @Flag var withSoftnet: Bool = false
+  @Flag(help: ArgumentHelp(visibility: .private))
+  var withSoftnet: Bool = false
```
fkorotkov, line 41: Maybe let's move the declaration after `netSoftnet` and add a `todo` to remove in `1.0.0`?

edigaryev: How about we have an issue for that instead of a static `TODO`: https://github.com/cirruslabs/tart/issues/276?

Sources/tart/Commands/Run.swift
edigaryev: I don't think that (2) it will be hard to remove that and (2) that we should remove it because `validate()`'s errors are based on the specific flag used.

fkorotkov (approved):

Issue #234, opened 2022-09-07
Tart registry operations speed


victor-yousician:
Our setup: m1 mac mini in MacStadium datacenter in Dublin on a 1 Gb/s connection (speedtest confirms that that's what we have).
Image size - 77.8 GB, not using the chunksize parameter
AWS ECR (Ireland) - push and pull max out at 19 MB/s and take over an hour.
ghcr.io - push 10 MB/s, takes over 2 hours, pull 51 MB/s - 25 minutes
Is this normal? Any way to improve that?

fkorotkov:
That's weird. I was able to reproduce a low speed of around 19 MB/s (160 Mpbs) in a similar setup where I had a Mac Mini on a symmetric 1 Gbps channel in a datacenter North America and was pushing to GitHub's registry.

I usually push images from home where I have an asymmetric 1 Gbps channel (only have 40 MB/s uplink). And when I push from home I always max out the uplink. It weird to see that I see slower push from a datacenter.

Me and @edigaryev were discussing to get rid of the overly complex https://github.com/swift-server/async-http-client which seems introducing more complexity then benefits... Especially when things like #168 happens.

edigaryev:
@victor-yousician can you check the speed against a server running on `localhost`? This will allow us to establish some more-or less reproduce baseline and, perhaps, solve your issue.

Here's an instruction on how to do that: https://github.com/cirruslabs/tart/issues/165#issuecomment-1213211307.

victor-yousician:
I built and ran a docker registry locally (without a container) on a Mac Studio (M1U/20/64/32/128GB/4TB/10G). Tested with the same image - 77.8 GB.
Push 138 MB/s - 9:36
Pull 340 MB/s - 3:54
Registry version
./registry github.com/distribution/distribution/v3 v2.7.0-2021-g78b9c98c

edigaryev:
@victor-yousician could you check if the Tart build in the #260 (https://cirrus-ci.com/task/6068291209789440) improves things for you?

Note that this binary won't be able to start a VM, only do basic operations like listing VMs and pulling/pushing them.

edigaryev:
@victor-yousician actually the #260 was just released [as a part of Tart 0.30.0 version](htps://github.com/cirruslabs/tart/releases/tag/0.30.0), please check it out.

Will close this for now.

victor-yousician:
local pull speed improved nicely - 483 MB/s (up from 340 MB/s)
[...]

PR #280, opened 2022-10-16
Downloads disk layers in parallel

### LINKED ISSUE #279: Poor performance pulling and pushing to ECR
Hi!

I'm not getting great performance with ECR. That is an issue for us because we are integrating tart into our macOS CI, which looks like this: pull an image, do things to it, and push the result back. As it stands, tart adds quite a lot of latency before tests can run, due to the slow pulling speed. Pushing is also an issue that prevents the jobs from moving on after the builds are finished.

I recently updated to release 0.30.0 hoping to improve this, so the numbers below are on this release.

I'm getting around 14MB/s when uploading, the image below is 27GB compressed and uploaded in 33min
```
> tart push tag ACCOUNT_ID.dkr.ecr.eu-west-1.amazonaws.com/macos/repo:tag
21:13:11.446744Z   pushing tag to ACCOUNT_ID.dkr.ecr.eu-west-1.amazonaws.com/macos/repo:tag...
21:13:11.479768Z   pushing config...
21:13:12.239447Z   pushing disk... this will take a while...
21:13:12.240589Z   0%

...

21:36:53.809777Z   100%
21:37:01.597372Z   pushing NVRAM...
21:37:03.378103Z   pushing manifest for tag...
```

Here is an example of pulling a 37.7GB image that took 1h20, so pulling is slower at 8MB/s.
```
> tart pull ACCOUNT_ID.dkr.ecr.eu-west-1.amazonaws.com/macos/repo:tag
15:20:29.398440Z   pulling ACCOUNT_ID.dkr.ecr.eu-west-1.amazonaws.com/macos/repo:tag...
15:20:29.398715Z   pulling manifest...
15:20:30.011137Z   pulling disk (37.7 GB compressed)...
15:20:30.011326Z   0%

...

16:42:32.350457Z   100%
16:42:32.381668Z   pulling NVRAM...
```

My setup is an AWS mac2.metal (M1 mac mini), hosted in the eu-west-1 region and working with ECR in that same region. The disk is a 500GB EBS gp2 volume which should provide 250 MB/s of throughput, and the network is a 10Gb/s link. 

I would appreciate any help to improve the performance. I'm also working with @saracen, who you've already been in touch with.

fkorotkov:
This change makes the `pull` to download layers in parallel into separate files. Than all the files are sequentially decompressed and are written into the resulting `disk.img`.

This change allowed to fully saturate a 1 Gbps connection while consuming around 30 Mb of RAM.

Fixes #279 #234 #283


fkorotkov:
@adrienkohlbecker will be interesting to see if you can check this binary https://api.cirrus-ci.com/v1/artifact/task/5985352270741504/binary/.build/debug/tart in your AWS setup.

fkorotkov:
On the second thought I think I'll try to follow up with a change to download layers in parallel into temporary files and then I archive them into the final disk file.

fkorotkov:
@adrienkohlbecker  we went with less invasive refactoring in #282 which showed the same results in our test environment with 1 Gbps connection. I winder if you can run some experiment in your 10 Gbps case. Could you please try a [binary from this PR](https://api.cirrus-ci.com/v1/artifact/task/5287316603273216/binary/.build/debug/tart) vs `0.32.0` version.

edigaryev:
This change gets the job done indeed, but is way too intrusive, as you've noted above.

What do you think about a version of this that would abstract all of the `URLSession.download()` details and file-related peculiarities away, essentially providing a `URLSession.bytes()` replacement, but faster (probably using a different return type for it's first argument, similarly to what's been done in https://github.com/cirruslabs/tart/pull/282).

Not only this will simplify the changeset and it's reviewal, but allow for easier "URL fetcher" mechanism swaps in the future, similarly to what's been already done with `Network` protocol.

fkorotkov:
Abstracted all the work with `URLSession` into static methods of a new class called `URLFetcher`.

Since #282 fixed the regression. Please look at this change as a further improvement and not a replacement for. This optimization of downloading in parallel to files on disk is a common pattern. This optimization IMO should show the best possible performance in terms of memory consumption and IO utilization.

edigaryev:
I've gave this one more shot today and still think this can be implemented way more cleanly by properly abstracting the `URLSession.shared.data()` and `URLSession.shared.download()` methods, as I've mentioned above.

In `VMDirectory+OCI.swift`, it seems that the code is actually re-reading the downloaded blobs into memory. What if we expose two methods in `Fetcher`:

* `fetch() -> (AsyncThrowingChannel<Data, Error>, HTTPURLResponse)`
  * calls `URLSession.shared.data()`
  * streams the received contents to the channel (once)
* `fetchViaFile() -> (AsyncThrowingChannel<Data, Error>, HTTPURLResponse)`
  * calls `URLSession.shared.download()`
  * opens `FileHandle` to the downloaded file
  * removes downloaded file
  * streams `FileHandle`'s contents to the channel (multiple times)

Then, in `Registry.authAwareRequest()`, introduce an argument `viaFile: Bool = false`, which will work like this:

```swift
if viaFile {
  return try await Fetcher().fetchViaFile(request)
}

return try await Fetcher().fetch(request)
```

Then, set `viaFile: true` in `Registry.pullBlob()` to only download blobs using files.

This abstraction would dramatically reduce the code changes needed to support parallel blob downloads via files, prevent repeatable authentication logic in `Registry` and allow us to more quickly switch to other method in the future, because we won't be tied to dealing with files.

PR #281, opened 2022-10-17
Relax status code acceptance

### LINKED ISSUE #273: Unexpected 201 when uploading to ECR with --chunk-size
Hi!

It looks like setting a `--chunk-size` is throwing errors when uploading to ECR. If I understood correctly, this should be set to more than 5 due to S3's multipart upload API limitations. Is there a value you recommend for best performance? 
In any case, it looks like the upload starts, some data is pushed, and then it receives an unexpected HTTP 201 and bails out. 

```
$ tart push --chunk-size 10 repo ACCOUNT_ID.dkr.ecr.eu-west-1.amazonaws.com/macos/repo:tag
pushing repo to ACCOUNT_ID.dkr.ecr.eu-west-1.amazonaws.com/macos/repo:tag...
pushing config...
pushing disk... this will take a while...
2%
UnexpectedHTTPStatusCode(when: "streaming blob to https://ACCOUNT_ID.dkr.ecr.eu-west-1.amazonaws.com/v2/macos/repo/blobs/uploads/b49173cd-0760-4e3b-9905-ba3cd0e42cca", code: 201, details: "")
```

fkorotkov:
Fixes #273

### DIFF
--- a/Sources/tart/OCI/Registry.swift
+++ b/Sources/tart/OCI/Registry.swift
@@ -214,6 +214,7 @@ class Registry {
         parameters: lastChunk ? ["digest": digest] : [:],
         body: chunk
       )
+      // always accept both statuses since AWS ECR is not following specification
       if response.statusCode != HTTPCode.Created.rawValue && response.statusCode != HTTPCode.Accepted.rawValue {
         throw RegistryError.UnexpectedHTTPStatusCode(when: "streaming blob to \(uploadLocation)",
           code: response.statusCode, details: data.asText())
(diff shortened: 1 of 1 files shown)


Sources/tart/OCI/Registry.swift
```diff
@@ -214,8 +214,7 @@ class Registry {
         parameters: lastChunk ? ["digest": digest] : [:],
         body: chunk
       )
-      let expectedStatus = lastChunk ? HTTPCode.Created.rawValue : HTTPCode.Accepted.rawValue
-      if response.statusCode != expectedStatus {
+      if response.statusCode != HTTPCode.Created.rawValue && response.statusCode != HTTPCode.Accepted.rawValue {
```
edigaryev, line 217: Do you think it would be good idea to leave a comment here since it's working according to a specification anymore?

edigaryev (approved):

PR #282, opened 2022-10-18
Use URLSession.dataTask() with delegate instead of URLSession.bytes()

edigaryev:
A simpler alternative to https://github.com/cirruslabs/tart/pull/280.

Resolves https://github.com/cirruslabs/tart/issues/279, https://github.com/cirruslabs/tart/issues/234.

### DIFF
--- a/Sources/tart/Fetcher.swift
+++ b/Sources/tart/Fetcher.swift
@@ -1,0 +1,46 @@
+import Foundation
+import AsyncAlgorithms
+
+class Fetcher: NSObject, URLSessionTaskDelegate, URLSessionDelegate, URLSessionDataDelegate {
+    let responseCh = AsyncThrowingChannel<URLResponse, Error>()
+    let dataCh = AsyncThrowingChannel<Data, Error>()
+
+    func fetch(_ request: URLRequest) async throws -> (AsyncThrowingChannel<Data, Error>, URLResponse) {
+        let session = URLSession(configuration: .default, delegate: self, delegateQueue: nil)
+        let task = session.dataTask(with: request)
+        task.resume()
+
+        // Wait for the response and only then return
+        var iter = responseCh.makeAsyncIterator()
+        let response = try await iter.next()!
+
+        return (dataCh, response)
+    }
+
+    func urlSession(_ session: URLSession, dataTask: URLSessionDataTask, didReceive response: URLResponse) async -> URLSession.ResponseDisposition {
+        await responseCh.send(response)
+
+        return .allow
+    }
+
+    func urlSession(_ session: URLSession, dataTask: URLSessionDataTask, didReceive data: Data) {
+        let sema = DispatchSemaphore(value: 0)
+
+        Task {
+            await dataCh.send(data)
+            sema.signal()
+        }
+
+        sema.wait()
+    }
+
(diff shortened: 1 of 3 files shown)


Sources/tart/Fetcher.swift
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    func urlSession(_ session: URLSession, dataTask: URLSessionDataTask, didReceive response: URLResponse) async -> URLSession.ResponseDisposition {
+        await responseCh.send(response)
+
+        return .allow
+    }
+
+    func urlSession(_ session: URLSession, dataTask: URLSessionDataTask, didReceive data: Data) {
+        let sema = DispatchSemaphore(value: 0)
```
fkorotkov, line 27: Why there is a need for a semaphore?

edigaryev: In order to wait for a `Task` to terminate, otherwise we will get results out-of-order. We also can't use `Task.result` because it's `async`, but we're in a `sync` context.

See also https://stackoverflow.com/a/42484670/9316533 and https://stackoverflow.com/a/47041695/9316533.

Sources/tart/Fetcher.swift
```diff
@@ -0,0 +1,46 @@
+import Foundation
+import AsyncAlgorithms
+
+class Fetcher: NSObject, URLSessionTaskDelegate, URLSessionDelegate, URLSessionDataDelegate {
+    let responseCh = AsyncThrowingChannel<URLResponse, Error>()
+    let dataCh = AsyncThrowingChannel<Data, Error>()
+
+    func fetch(_ request: URLRequest) async throws -> (AsyncThrowingChannel<Data, Error>, URLResponse) {
+        let session = URLSession(configuration: .default, delegate: self, delegateQueue: nil)
```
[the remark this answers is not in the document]
edigaryev: Changed to `URLSession.shared` in https://github.com/cirruslabs/tart/pull/282/commits/4dd7f612bf6712500a52fb4e56dca97837e6d34b.

Sources/tart/Fetcher.swift
```diff
@@ -0,0 +1,46 @@
+import Foundation
+import AsyncAlgorithms
+
+class Fetcher: NSObject, URLSessionTaskDelegate, URLSessionDelegate, URLSessionDataDelegate {
```
fkorotkov, line 4: From the name it seems it can be used for fetching multiple requests but I don't think `fetch` can handle it.

edigaryev: Do you want to rename it? Or add an exception in `fetch()` if it was called twice?

Sources/tart/Fetcher.swift
```diff
@@ -0,0 +1,46 @@
...
+        await responseCh.send(response)
+
+        return .allow
+    }
+
+    func urlSession(_ session: URLSession, dataTask: URLSessionDataTask, didReceive data: Data) {
+        let sema = DispatchSemaphore(value: 0)
+
+        Task {
+            await dataCh.send(data)
```
fkorotkov, line 30: Wait. It seems it will wait here until the data is consumed?

fkorotkov (approved):
I don't like `Fetcher` implementaion but let's unblock folks and then re-iterate.

PR #284, opened 2022-10-18
OCI: always read channel until end

### LINKED ISSUE #283: Pull hangs in 0.32.0
The message I get on `^C`
```
% tart pull localhost/monterey_x13-3-1_a30_u2020-3-34f1:20221014 
pulling localhost/monterey_x13-3-1_a30_u2020-3-34f1:20221014...
pulling manifest...
^Ctart/Fetcher.swift:15: Fatal error: Unexpectedly found nil while unwrapping an Optional value
zsh: trace trap  tart pull localhost/monterey_x13-3-1_a30_u2020-3-34f1:20221014
```

edigaryev:
Resolves #283.

### DIFF
--- a/Sources/tart/OCI/Registry.swift
+++ b/Sources/tart/OCI/Registry.swift
@@ -295,6 +295,7 @@ class Registry {
     var (channel, response) = try await authAwareRequest(request: request)
 
     if doAuth && response.statusCode == HTTPCode.Unauthorized.rawValue {
+      _ = try await channel.asData()
       try await auth(response: response)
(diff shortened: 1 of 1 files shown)


fkorotkov:
I"m more and more in favor of going with #280 even though it's a bigger change. Handling these raw requests seems to be a problem given that the source code is not available. IMO just relying on the `URLSession.download` and working with files is fine.

fkorotkov (approved):
Let's make `0.32.1` out there. But will strongly consider to go with #280

PR #306, opened 2022-11-07
OCI: pull blobs via file

edigaryev:
A PoC for https://github.com/cirruslabs/tart/pull/280#issuecomment-1300105314.

### DIFF
--- a/Sources/tart/Fetcher.swift
+++ b/Sources/tart/Fetcher.swift
@@ -2,7 +2,15 @@ import Foundation
 import AsyncAlgorithms
 
 class Fetcher: NSObject, URLSessionTaskDelegate, URLSessionDelegate, URLSessionDataDelegate {
-  func fetch(_ request: URLRequest) async throws -> (AsyncThrowingChannel<Data, Error>, HTTPURLResponse) {
+  func fetch(_ request: URLRequest, viaFile: Bool = false) async throws -> (AsyncThrowingChannel<Data, Error>, HTTPURLResponse) {
+    if viaFile {
+      return try await fetchViaFile(request)
+    }
+
+    return try await fetchViaMemory(request)
+  }
+
+  private func fetchViaMemory(_ request: URLRequest) async throws -> (AsyncThrowingChannel<Data, Error>, HTTPURLResponse) {
     let dataCh = AsyncThrowingChannel<Data, Error>()
 
     let (data, response) = try await URLSession.shared.data(for: request)
@@ -16,7 +24,7 @@ class Fetcher: NSObject, URLSessionTaskDelegate, URLSessionDelegate, URLSessionD
     return (dataCh, response as! HTTPURLResponse)
   }
 
-  func fetchViaFile(_ request: URLRequest) async throws -> (AsyncThrowingChannel<Data, Error>, HTTPURLResponse) {
+  private func fetchViaFile(_ request: URLRequest) async throws -> (AsyncThrowingChannel<Data, Error>, HTTPURLResponse) {
     let dataCh = AsyncThrowingChannel<Data, Error>()
 
     let (fileURL, response) = try await URLSession.shared.download(for: request)
--- a/Sources/tart/Fetcher.swift
+++ b/Sources/tart/Fetcher.swift
@@ -20,6 +20,11 @@ class Fetcher: NSObject, URLSessionTaskDelegate, URLSessionDelegate, URLSessionD
     let dataCh = AsyncThrowingChannel<Data, Error>()
 
     let (fileURL, response) = try await URLSession.shared.download(for: request)
+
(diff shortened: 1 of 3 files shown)


Sources/tart/Fetcher.swift
```diff
@@ -2,45 +2,35 @@ import Foundation
...
-        return .allow
-    }
+  func fetchViaFile(_ request: URLRequest) async throws -> (AsyncThrowingChannel<Data, Error>, HTTPURLResponse) {
+    let dataCh = AsyncThrowingChannel<Data, Error>()
 
-    func urlSession(_ session: URLSession, dataTask: URLSessionDataTask, didReceive data: Data) {
-        let sema = DispatchSemaphore(value: 0)
+    let (fileURL, response) = try await URLSession.shared.download(for: request)
+    let fh = try FileHandle(forReadingFrom: fileURL)
+    try FileManager.default.removeItem(at: fileURL)
```
fkorotkov, line 24: Let's add a comment explaining the magic of acquiring a file handle and then removing the file but still being able to read it.

Sources/tart/Fetcher.swift
```diff
@@ -2,45 +2,35 @@ import Foundation
...
 class Fetcher: NSObject, URLSessionTaskDelegate, URLSessionDelegate, URLSessionDataDelegate {
-    let responseCh = AsyncThrowingChannel<URLResponse, Error>()
+  func fetch(_ request: URLRequest) async throws -> (AsyncThrowingChannel<Data, Error>, HTTPURLResponse) {
     let dataCh = AsyncThrowingChannel<Data, Error>()
 
-    func fetch(_ request: URLRequest) async throws -> (AsyncThrowingChannel<Data, Error>, URLResponse) {
-        let task = URLSession.shared.dataTask(with: request)
-        task.delegate = self
-        task.resume()
+    let (data, response) = try await URLSession.shared.data(for: request)
```
fkorotkov, line 8: I wonder if that needs to be inside the `Task` 🤔  But probably it's better to keep it here for better error handling and stacktrace understanding.

Sources/tart/Fetcher.swift
```diff
@@ -2,45 +2,35 @@ import Foundation
 import AsyncAlgorithms
 
 class Fetcher: NSObject, URLSessionTaskDelegate, URLSessionDelegate, URLSessionDataDelegate {
```
fkorotkov, line 4: What do you think of making `Fetcher` fileprivate here and introducing public `URLFetcher` class with static `fetch` method that will take `viaFile` argument.

This way we'll abstract things even more and will hide and limit the complexity of `Fetcher` to this file (e.g. prevent from initializing `Fetcher` somewhere and calling methods twice).

Sources/tart/Fetcher.swift
edigaryev: See https://github.com/cirruslabs/tart/pull/306/commits/9363c7cfd7131d8a51fd02c2de96f37984c5246d.

Sources/tart/Fetcher.swift
edigaryev: Since we're returning `URLResponse` immediately, it makes sense to only process `Data` inside of a `Task`.

Sources/tart/Fetcher.swift
edigaryev: See https://github.com/cirruslabs/tart/pull/306/commits/cf180458e371fb69eeec3ba0060d8a2336d428d4.

fkorotkov: What do you think about the `fileprivate` part and having URLFetcher with only one method?

edigaryev: > What do you think about the `fileprivate` part and having URLFetcher with only one method?

I fail to see the advantages of this approach, compared to the solution in https://github.com/cirruslabs/tart/commit/cf180458e371fb69eeec3ba0060d8a2336d428d4. Can you explain?

With https://github.com/cirruslabs/tart/commit/cf180458e371fb69eeec3ba0060d8a2336d428d4, the `Fetcher` also has only one publicly-accessible method and this approach doesn't introduce extraneous classes that might confuse the consumer.

fkorotkov: `fileprivate` won't be visible to users. Right now one can create a single `Fetcher` instance and try to use it twice.

edigaryev: > Right now one can create a single `Fetcher` instance and try to use it twice.

That's actually not an issue any more since there's no shared state stored between `fetch()` invocations.

I've outlined this in https://github.com/cirruslabs/tart/pull/306/commits/2704cbe384f0a0e85f4da5304ede847bc671ac5e by making the method `static`.

fkorotkov (approved):

Issue #302, opened 2022-11-04
tart get command


tomjn:
We can use `tart set` to change a VMs cpu/memory/etc but there is no way to retrieve this data, a `tart get` command would be useful

PR #309, opened 2022-11-08
Implement Get command

evanburkeyhl:
This PR implements a get command as outlined in https://github.com/cirruslabs/tart/issues/302. The output was formatted to look similar to other `tart` commands and is as follows:

```
> tart get m
cpu     memory  disk    display
4       8192MB  20GB    1024x768
```

### DIFF
--- a/Sources/tart/Commands/Get.swift
+++ b/Sources/tart/Commands/Get.swift
@@ -13,12 +13,13 @@ struct Get: AsyncParsableCommand {
       let vmConfig = try VMConfig(fromURL: vmDir.configURL)
       let diskSize = try vmDir.sizeBytes() / 1024 / 1024 / 1000
 
-      print("""
-            cpu cores: \(vmConfig.cpuCount)
-            memory:    \(vmConfig.memorySize / 1024 / 1024)
-            disk size: \(diskSize) GB
-            display:   \(vmConfig.display.width) x \(vmConfig.display.height)
-            """)
+      print("cpu\tmemory\tdisk\tdisplay")
+
+      var s = "\(vmConfig.cpuCount)\t"
+      s += "\(vmConfig.memorySize / 1024 / 1024)MB\t"
+      s += "\(diskSize)GB\t"
+      s += "\(vmConfig.display.width)x\(vmConfig.display.height)"
+      print(s)
 
       Foundation.exit(0)
     } catch {
--- a/Sources/tart/Commands/Get.swift
+++ b/Sources/tart/Commands/Get.swift
@@ -13,7 +13,7 @@ struct Get: AsyncParsableCommand {
       let vmConfig = try VMConfig(fromURL: vmDir.configURL)
       let diskSize = try vmDir.sizeBytes() / 1024 / 1024 / 1000
 
-      print("cpu\tmemory\tdisk\tdisplay")
+      print("CPU\tMemory\tDisk\tDisplay")
 
       var s = "\(vmConfig.cpuCount)\t"
       s += "\(vmConfig.memorySize / 1024 / 1024)MB\t"
--- a/Sources/tart/Commands/Get.swift
+++ b/Sources/tart/Commands/Get.swift
@@ -1,0 +1,30 @@
+import ArgumentParser
+import Foundation
+
+struct Get: AsyncParsableCommand {
+  static var configuration = CommandConfiguration(commandName: "get", abstract: "Get a VM's configuration")
+
+  @Argument(help: "VM name")
+  var name: String
+
+  func run() async throws {
+    do {
+      let vmDir = try VMStorageLocal().open(name)
+      let vmConfig = try VMConfig(fromURL: vmDir.configURL)
+      let diskSize = try vmDir.sizeBytes() / 1024 / 1024 / 1000
+
+      print("""
+            cpu cores: \(vmConfig.cpuCount)
+            memory:    \(vmConfig.memorySize / 1024 / 1024)
+            disk size: \(diskSize) GB
+            display:   \(vmConfig.display.width) x \(vmConfig.display.height)
+            """)
+
+      Foundation.exit(0)
+    } catch {
+      print(error)
+
+      Foundation.exit(1)
+    }
+  }
+}
--- a/Sources/tart/Commands/Get.swift
+++ b/Sources/tart/Commands/Get.swift
@@ -11,13 +11,13 @@ struct Get: AsyncParsableCommand {
     do {
       let vmDir = try VMStorageLocal().open(name)
       let vmConfig = try VMConfig(fromURL: vmDir.configURL)
-      let diskSize = try vmDir.sizeBytes() / 1024 / 1024 / 1000
+      let diskSize = try vmDir.sizeBytes() / 1000 / 1000 / 1000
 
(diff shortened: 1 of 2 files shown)


Sources/tart/Commands/Get.swift
```diff
@@ -0,0 +1,31 @@
...
+  @Argument(help: "VM name")
+  var name: String
+
+  func run() async throws {
+    do {
+      let vmDir = try VMStorageLocal().open(name)
+      let vmConfig = try VMConfig(fromURL: vmDir.configURL)
+      let diskSize = try vmDir.sizeBytes() / 1024 / 1024 / 1000
+
+      print("cpu\tmemory\tdisk\tdisplay")
```
edigaryev, line 16: Since this is a human-readable output, what do you think about capitalizing the row names, to be consistent with the `tart list`'s output?

```suggestion
      print("CPU\tMemory\tDisk\tDisplay")
```

Sources/tart/Commands/Get.swift
```diff
@@ -0,0 +1,31 @@
...
+  static var configuration = CommandConfiguration(commandName: "get", abstract: "Get a VM's configuration")
+
+  @Argument(help: "VM name")
+  var name: String
+
+  func run() async throws {
+    do {
+      let vmDir = try VMStorageLocal().open(name)
+      let vmConfig = try VMConfig(fromURL: vmDir.configURL)
+      let diskSize = try vmDir.sizeBytes() / 1024 / 1024 / 1000
```
edigaryev, line 14: That calculation looks a bit weird to me, why both 1024 and 1000? 🤔 

https://github.com/cirruslabs/tart/blob/555715588d4794fa0875d8dc3998bda0099bd77d/Sources/tart/VMDirectory.swift#L85

What do you think about showing the output in GiB's instead?

```suggestion
      let diskSize = try vmDir.sizeBytes() / 1000 / 1000 / 1000
```

Sources/tart/Commands/Get.swift
```diff
@@ -0,0 +1,31 @@
...
+
+      print("cpu\tmemory\tdisk\tdisplay")
+
+      var s = "\(vmConfig.cpuCount)\t"
+      s += "\(vmConfig.memorySize / 1024 / 1024)MB\t"
+      s += "\(diskSize)GB\t"
```
edigaryev, line 20: See https://github.com/cirruslabs/tart/pull/309/files#r1017574000.

```suggestion
      s += "\(diskSize) GiB\t"
```

Sources/tart/Commands/Get.swift
```diff
@@ -0,0 +1,31 @@
...
+  func run() async throws {
+    do {
+      let vmDir = try VMStorageLocal().open(name)
+      let vmConfig = try VMConfig(fromURL: vmDir.configURL)
+      let diskSize = try vmDir.sizeBytes() / 1024 / 1024 / 1000
+
+      print("cpu\tmemory\tdisk\tdisplay")
+
+      var s = "\(vmConfig.cpuCount)\t"
+      s += "\(vmConfig.memorySize / 1024 / 1024)MB\t"
```
edigaryev, line 19: Added a spacing to improve readability:

```suggestion
      s += "\(vmConfig.memorySize / 1024 / 1024) MB\t"
```

Sources/tart/Commands/Get.swift
fkorotkov: macOS doesn't use `GiB` in the UI even though `GB` is `GiB`. See https://github.com/cirruslabs/tart/blob/8961c5189a792cda828b8e43dade23c38bae5736/Sources/tart/VMDirectory.swift#L84-L85

So to be consistent with macOS Finder we should IMO stick with GB.

fkorotkov:
@tomjn is it how you envisioned `get` command? Or you wanted an ability to get each particular value with flags like `--memory`, etc.?

tomjn:
Individual flags would be useful for bash scripting but not essential. From what I can see this is output is essentially a TSV? It can be parsed to retrieve values but those are still key/value pairs

tomjn:
If you were to go with individual flags, it may be better to follow other CLI's as a general `get` and `set` much like `git config`, this also provides futureproofing for additional values without adding a huge number of parameters. e.g. `tart get memory` or `tart set memory 4096mb`. I believe these are alls tored in a JSON file in the Tart VMs folder?

Reading through the PR I'm also unsure what to expect for a VM created with no display, do all Tart VMs have a default screen resolution even if they're ran headless? For my usage it's mainly academic as the parameter is passed during the `tart run` command anyway

Sources/tart/Commands/Get.swift
```diff
@@ -0,0 +1,31 @@
...
+struct Get: AsyncParsableCommand {
+  static var configuration = CommandConfiguration(commandName: "get", abstract: "Get a VM's configuration")
+
+  @Argument(help: "VM name")
+  var name: String
+
+  func run() async throws {
+    do {
+      let vmDir = try VMStorageLocal().open(name)
+      let vmConfig = try VMConfig(fromURL: vmDir.configURL)
```
tomjn, line 13: what happens if I pass a `name` that doesn't exist `tart get sepllingmistake`

tomjn: would that bubble up via the `throws` keyword? Apologies if my Swift knowledge isn't up to scratch

evanburkeyhl: It does bubble up but the error message is somewhat cryptic. 

```
> ./tart get masdf
UninitializedVMDirectoryError()
```

However, this is in line with what happens from an incorrect name using `tart set`, which is why I did not modify it

Sources/tart/Commands/Get.swift
evanburkeyhl: I think that's a good idea

Sources/tart/Commands/Get.swift
evanburkeyhl: In retrospect that was a weird way to show GB. I like the idea of GiB instead, it's more accurate

evanburkeyhl:
Thanks for taking a look. The macOS team at @huntresslabs use tart on a daily basis and we'd like to start contributing back to the project when we have time, this PR was me testing the waters a bit.

edigaryev (approved):

fkorotkov (approved):

fkorotkov:
Hey @evanburkeyhl,

We are preparing 1.0.0 release of Tart at some point in January and would like to clean things up and stabilize the API to provide long term support. Therefore we are introducing Contributor License Agreement which is common practice among mature OSS projects.

fkorotkov:
Hey @evanburkeyhl,

We are preparing 1.0.0 release of Tart at some point in January and would like to clean things up and stabilize the API to provide long term support. Therefore we are introducing Contributor License Agreement which is common practice among mature OSS projects.

evanburkeyhl:
@fkorotkov I am reviewing the CLA with our legal department and will get back to you after discussing with them.

evanburkeyhl:
@fkorotkov Our legal department has given their blessing and I have signed the agreement. Let me know if there's anything else you need

fkorotkov:
@evanburkeyhl appreciate the quick turn around! Thank you!

PR #317, opened 2022-11-10
Prefer official Homebrew distribution

umireon:
I've added tart to the official Homebrew repository.
https://github.com/Homebrew/homebrew-core/pull/114571

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ Built by CI engineers for your automation needs. Here are some highlights of Tar
 Try running a Tart VM on your Apple Silicon device running macOS Monterey or later (will download a 25 GB image):
 
 ```shell
-brew install cirruslabs/cli/tart
+brew install tart
 tart clone ghcr.io/cirruslabs/macos-ventura-base:latest ventura-base
 tart run ventura-base
 ```


edigaryev:
It's nice to see Tart being available in Homebrew directly, but this is somewhat a regression due to slower release time, the Homebrew's repository and review process being the bottleneck here.

Perhaps leaving both options to install Tart in `README.md` and some explanation on how to choose between the two would be a nice trade-off here.

edigaryev:
> Especially for version updates, they have update detectors and maintainers and contributors are eager to keep every formula up to date.

Good to know that!

But still, we need to keep in mind that it's not immediate from Tart's side, like it is now. One has to do two things now, instead of one. I wonder if this can be automated somehow.

fkorotkov:
Appreciate the effort of putting out `tart` into the official Homebrew repository! I wish you would've create an discussion first though!

Two things will not work for Tart here. First, we do need signing with our certificate to get the elevated network entitlements. But most importantly we need signing with some valid developer certificate (and not self-signed) so users don't need to allow access to Keychain every time you update Tart.

On top of the things, we need an automated way to propagate updates upon release. Right now [we use `Goreleaser`](https://github.com/cirruslabs/tart/blob/f37372da2850acb51690f6f212ae8faaa841f400/.goreleaser.yml#L33-L53). What's the story of automatically creating PR for updates to Homebrew core?

fkorotkov:
Closing it since it's unclear how to support the signing part since it requires private keys from our side. Either the homebrew formula should download prebuilt binaries or we should just remove to lower the confusion.

Issue #214, opened 2022-08-30
Option to provide registry credentials via environment variables


fkorotkov:
In an automated environment like CI especially if it's not ephemeral Keychain is not a good option for storing credentials. One will always need to clean the credentials. Credential helpers #167 also has an intention to store things locally.

IMO for one time pushes to registry Tart should have an option to provide credentials via `TART_REGISTRY_USERNAME` and `TART_REGISTRY_PASSWORD`.

mortidlo:
This feature would be really helpful for our project. We use tart in GitLab CI jobs and we have to work with keychains. Using env variables for password/username would be great.

edigaryev:
> IMO for one time pushes to registry Tart should have an option to provide credentials via `TART_REGISTRY_USERNAME` and `TART_REGISTRY_PASSWORD`.

I think this will only support a single registry **or** give out the registry-specific credentials to all registries that Tart connects to, which is not ideal.

> This feature would be really helpful for our project. We use tart in GitLab CI jobs and we have to work with keychains. Using env variables for password/username would be great.

@mortidlo can you elaborate a bit more on the problem you're facing with Tart's usage of Keychain? Is it the Keychain cleaning process, or something else?

edigaryev:
> So for us the most convenient way would be setup environment variables for authentication same way as for example docker does for this one time pushes. So we would just type `tart login ….` inside of job and then push image.

Thanks for the explanation, do you have some link to the documentation about Docker one time pushes?

Also, does:

```
echo -n $REGISTRY_PASSWORD | tart login --username $REGISTRY_USERNAME --password-stdin
```

...work for you?

mortidlo:
Yes login works exactly this way. I think you only need to learn tart store credentials outside of keychain in something similar to .docker.json and use Bearer authentication against registry endpoint.

Edit: I saw you have already merged in #320 and also documented it. Thank you very much.

fkorotkov:
Not sure why this issue was not automatically closed though.

PR #320, opened 2022-11-10
Option to provide registry credentials via environment variables

### LINKED ISSUE #124: Add Xcode version to --version
Since the underlying `Virtualization.Framework` is changing. I think it will make sense to add that info to `tart --version`.

fkorotkov:
Fixes #124

### DIFF
--- a/Sources/tart/Credentials/EnvironmentCredentialsProvider.swift
+++ b/Sources/tart/Credentials/EnvironmentCredentialsProvider.swift
@@ -1,0 +1,15 @@
+import Foundation
+
+class EnvironmentCredentialsProvider: CredentialsProvider {
+  func retrieve(host: String) throws -> (String, String)? {
+    let username = ProcessInfo.processInfo.environment["TART_REGISTRY_USERNAME"]
+    let password = ProcessInfo.processInfo.environment["TART_REGISTRY_PASSWORD"]
+    if username != nil && password != nil {
+      return (username!, password!)
+    }
+    return nil
+  }
+
+  func store(host: String, user: String, password: String) throws {
+  }
+}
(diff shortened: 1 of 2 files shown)


Sources/tart/Credentials/EnvironmentCredentialsProvider.swift
```diff
@@ -0,0 +1,15 @@
+import Foundation
+
+class EnvironmentCredentialsProvider: CredentialsProvider {
+  func retrieve(host: String) throws -> (String, String)? {
+    let username = ProcessInfo.processInfo.environment["TART_REGISTRY_USERNAME"]
+    let password = ProcessInfo.processInfo.environment["TART_REGISTRY_PASSWORD"]
+    if username != nil && password != nil {
```
edigaryev, line 7:
```suggestion
if let username = username, let password = password {
```

Sources/tart/Credentials/EnvironmentCredentialsProvider.swift
```diff
@@ -0,0 +1,15 @@
...
+      return (username!, password!)
```
edigaryev, line 8:
```suggestion
    if let username = username, let password = password {
      return (username, password)
```

edigaryev:
What about the concern raised in https://github.com/cirruslabs/tart/issues/214#issuecomment-1309849274? It seems to be really easy to send credentials to another registry with this approach.

Also, these new environment variables need some documentation in the `README.md`.

fkorotkov:
I've documented both credential helpers and the env variables in [c2e9dc8](https://github.com/cirruslabs/tart/pull/320/commits/c2e9dc8b700b55581d942f86fbd8994002922afd).

Even though one can define `TART_REGISTRY_USERNAME` and `TART_REGISTRY_PASSWORD` in their `zsh` profile or something, the main use case is integrating with CI's secret management for easier pushes:

```yaml
task:
  env:
    TART_REGISTRY_USERNAME: "_json"
    TART_REGISTRY_PASWORD: ENCRYPTED[...]
  build_script: ...
  push_script: tart push ...
```

edigaryev (approved):

PR #328, opened 2022-11-16
Added Code Owners

fkorotkov:
So before auto-merging there is a need for approval from both of me and @edigaryev


.github/gi
```diff
@@ -0,0 +1 @@
+* @edigaryev @fkorotkov
```
edigaryev, line 1: Shouldn't we name this file `CODEOWNERS` instead, as per [GitHub's documentation](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#codeowners-file-location)?

edigaryev (approved):

Issue #323, opened 2022-11-15
Stuck if attached disk image is already mounted on host


gyfelton:
There is no error message indicating the instance stuck at launch because the disk is already mounted. To repro, do `tart run <some_vm> --disk <some_disk_path>` while having the disk image already mounted on host. VM will stuck at black screen with no error message when I expect some error showing up immediately and exit just like when we try to load a not supported disk image type.

Definitely just an enhancement but if anyone else have time wasted because of not realizing this, vote on this issue so we can find time to introduce a fix for it.

fkorotkov:
Thanks for reporting! We'll need to investigate if we can easily check whether a disk is in use already. 👌

testingbot:
When we specify a disk as `:ro` (readonly), the [check](https://github.com/cirruslabs/tart/commit/14cbc727adafb0449f54da71ddb580bbc592bf38) is probably not necessary?

edigaryev:
> When we specify a disk as `:ro` (readonly), the [check](https://github.com/cirruslabs/tart/commit/14cbc727adafb0449f54da71ddb580bbc592bf38) is probably not necessary?

Thanks for spotting this! I've addressed this in a separate PR: https://github.com/cirruslabs/tart/pull/334.

PR #333, opened 2022-11-21
tart run: check if the disks are locked

edigaryev:
It seems that the `Virtualization.Framework` tries to acquire a lock on the disk passed by us and hangs.

One obvious thing we can do is to check the lock before starting the actual VM and bail out if the lock is acquired by someone else.

Resolves https://github.com/cirruslabs/tart/issues/323.


fkorotkov (approved):

gyfelton:
thx for this quick fix!

Issue #301, opened 2022-11-04
Linux Examples


tomjn:
It would be helpful to have more Linux examples in the readme, _in particular_ examples of mounting folders into specific locations, as well as Linux VMs that aren't created from scratch

edigaryev:
> as well as Linux VMs that aren't created from scratch

Can you clarify which examples you'd like to see? Currently, the process for Linux VMs is the same as with macOS VMs.

tomjn:
Examples that include mounted folders specifically, but also to ensure that every MacOS example has a linux equivalent, even if it's just to demonstrate that they're the same

tomjn:
Also, arbitrary mounts. For example by default Vagrant mounts the folder with the vagrantfile to `/vagrant`, I'm unsure how I would achieve something similar with this on Linux without resorting to hardlinks/symlinks to `/mnt/shared/vagrant`. It's likely manual rather than auto-mounting will be needed for that

edigaryev:
> Examples that include mounted folders specifically

This will be addressed by #338.

>but also to ensure that every MacOS example has a linux equivalent, even if it's just to demonstrate that they're the same

Seems very redundant. It would probably would be better to stick with the common sense notion that unless documented, the instructions are the same for both macOS and Linux. What do you think?

>Also, arbitrary mounts. For example by default Vagrant mounts the folder with the vagrantfile to `/vagrant`, I'm unsure how I would achieve something similar with this on Linux without resorting to hardlinks/symlinks to `/mnt/shared/vagrant`. It's likely manual rather than auto-mounting will be needed for that

Most likely using the hard/soft links, yes. Otherwise this seems like an out of scope piece for the documentation, because it does not involve the Tart directly.

tomjn:
> Seems very redundant. It would probably would be better to stick with the common sense notion that unless documented, the instructions are the same for both macOS and Linux. What do you think?

Unless someone actually says they are the same this introduces ambiguity, e.g. are they the same, or is it that those parameters aren't supported on Linux, or is it that they're different but undocumented? Not everybody will interpret things the same, and if something is only mentioned under a MacOS heading it's heavily implied it's a MacOS only feature.

If instead it's a generic section that talks about adding folders to a guest, rather than separate duplicates that are MacOS and Linux specific, then that also does the job without being unnecessarily verbose

> Most likely using the hard/soft links, yes. Otherwise this seems like an out of scope piece for the documentation, because it does not involve the Tart directly.

[...]

PR #338, opened 2022-11-29
README.md: document --dir

edigaryev:
Resolves https://github.com/cirruslabs/tart/issues/301.

### DIFF
--- a/README.md
+++ b/README.md
@@ -278,7 +278,7 @@ This invocation calls the `tart pull` implicitly (if the image is not being pres
 
 To mount a directory, run the VM with the `--dir` argument:
 
-```
+```sh
 tart run --dir=project:~/src/project vm
 ```
 
--- a/README.md
+++ b/README.md
@@ -290,6 +290,14 @@ It is also possible to mount directories in read-only mode by adding a third par
 tart run --dir=project:~/src/project:ro vm
 ```
 
+To mount multiple directories, repeat the `--dir` argument for each directory:
+
+```sh
+tart run --dir=www1:~/project1/www --dir=www2:~/project2/www
+```
+
+Note that the first parameter in each `--dir` argument must be unique, otherwise only the last `--dir` argument using that name will be used.
+
 Note: to use the directory mounting feature, the host needs to run macOS 13.0 (Ventura) or newer.
 
 #### Accessing mounted directories in macOS guests
--- a/README.md
+++ b/README.md
@@ -286,7 +286,7 @@ Here, the `project` specifies a mount name, whereas the `~/src/project` is a pat
 
 It is also possible to mount directories in read-only mode by adding a third parameter, `ro`:
 
-```
+```sh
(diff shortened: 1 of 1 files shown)


fkorotkov (dismissed):

README.md
```diff
@@ -274,6 +274,42 @@ tart clone acme.io/remoteorg/name:latest my-local-vm-name
...
+
+#### Accessing mounted directories in Linux guests
+
+To be able to access the shared directories from the Linux guest, you need to manually mount the virtual filesystem first:
+
+```
+mount -t virtiofs com.apple.virtio-fs.automount /mnt/shared
+```
+
+The directory we've mounted above will be accessible from the `/mnt/shared/project` path inside a guest VM.
```
tomjn, line 311: does this mean that it's not possible to mount it to an arbitrary location, and that it must have the same name and be unique?

E.g. if I were to try to mount 2 folders named `public_html` in different locations that are not `/mnt/shared` would that be possible?

README.md
```diff
@@ -274,6 +274,42 @@ tart clone acme.io/remoteorg/name:latest my-local-vm-name
...
+
+The directory we've mounted above will be accessible from the `/Volumes/My Shared Files/project` path inside a guest VM.
+
+Note: to use the directory mounting feature, the guest VM needs to run macOS 13.0 (Ventura) or newer.
+
+#### Accessing mounted directories in Linux guests
+
+To be able to access the shared directories from the Linux guest, you need to manually mount the virtual filesystem first:
+
+```
```
tomjn, line 307:
```suggestion
```sh
```

README.md
```diff
@@ -274,6 +274,42 @@ tart clone acme.io/remoteorg/name:latest my-local-vm-name
...
+tart run --dir=project:~/src/project vm
+```
+
+Here, the `project` specifies a mount name, whereas the `~/src/project` is a path to the host's directory to expose to the VM.
+
+It is also possible to mount directories in read-only mode by adding a third parameter, `ro`:
+
+```
```
tomjn, line 289:
```suggestion
```sh
```

README.md
edigaryev: > E.g. if I were to try to mount 2 folders named `public_html` in different locations that are not `/mnt/shared` would that be possible?

I don't think so, you can only mount the "outer" directory that contains all of the "inner" mounts specified by `--dir`.

edigaryev: No, you can use `--dir` multiple times, this is explicitly stated in`tart run --help`.

tomjn: can that be included in the readme with an example? How does it handle when multiple dir are added but they have the same folder name e.g. `~/project1/www`  and `/project2/www`?

edigaryev: > can that be included in the readme with an example?

See https://github.com/cirruslabs/tart/pull/338/commits/7a26b3b4781288d9a7dd859ec97bce01b56a957e.

fkorotkov (approved):

PR #342, opened 2022-11-29
Fix SwiftFormat's URL/glob mishandling

edigaryev:
To fix the build on `main`.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,3 +1,5 @@
+use_compute_credits: true
+
 task:
   name: Test on Ventura
   alias: test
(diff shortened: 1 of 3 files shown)


fkorotkov (dismissed):
Weird it didn't error about missing config file 🤷‍♂️

edigaryev:
Seems like this is not going to work, a fix in the SwiftFormat is needed: https://github.com/nicklockwood/SwiftFormat/pull/1317.

I've pointed our `Package.swift` to a Git SHA with the fix, so it should be working before we get the fix from the upstream.

edigaryev:
Bumped the Git SHA to accomodate another fix needed: https://github.com/nicklockwood/SwiftFormat/pull/1317#issuecomment-1331381997.

fkorotkov (approved):

PR #343, opened 2022-12-01
Don't swallow VZVirtualMachine.start() exceptions

edigaryev:
Without this fix, we miss all of the errors thrown from `VZVirtualMachine.start()`, for example:

```
Error Domain=VZErrorDomain Code=6 "The maximum supported number of active virtual machines has been reached." UserInfo={NSLocalizedFailure=The number of virtual machines exceeds the limit., NSLocalizedFailureReason=The maximum supported number of active virtual machines has been reached.}
```

The lack of this particular error complicates problem diagnostics when running multiple VMs on a single host.


Sources/tart/VM.swift
```diff
@@ -219,21 +219,26 @@ class VM: NSObject, VZVirtualMachineDelegate, ObservableObject {
   func run(_ recovery: Bool) async throws {
     try network.run(sema)
 
-    DispatchQueue.main.sync {
+    let task = DispatchQueue.main.sync {
```
fkorotkov, line 222: What do you think of renaming it to something like `vmStartTask` or `startTask`?

edigaryev: See https://github.com/cirruslabs/tart/pull/343/commits/213d5cc4d98b4f279080c677ee8b9867bd3d1fc2.

fkorotkov (approved):

Issue #299, opened 2022-11-04
Filter tart list


tomjn:
Tart list currently displays both local and OCI:

```swift
      displayTable("local", try VMStorageLocal().list())
      displayTable("oci", try VMStorageOCI().list().map { (name, vmDir, _) in (name, vmDir) })
```

It should be possible to filter to show one or the other if desired

edigaryev:
Resolved by https://github.com/cirruslabs/tart/pull/349.

PR #349, opened 2022-12-14
Filter tart list by source

marc-48k:
An attempt to solve #299 by providing a `--source` option to the `list` command.

Some example output:

```shell
tart list --help                

OVERVIEW: List created VMs

USAGE: tart list [--quiet] [--source <source>]

OPTIONS:
  -q, --quiet             Only display VM names.
  --source <source>       Only display VMs from the specified source (e.g. --source local, --source oci).
  --version               Show the version.
  -h, --help              Show help information.
```

```shell
$ tart list
Source	Name
local 	monterey-base
local 	ventura-base
local 	ventura-new
oci   	ghcr.io/cirruslabs/macos-ventura-base:latest
```

```shell
tart list --source oci
Source  Name
oci   	ghcr.io/cirruslabs/macos-ventura-base:latest
```

```shell
tart list --source foo
Unknown source: 'foo'
```

```shell
tart list --source local --quiet
monterey-base
ventura-base
ventura-new
```

I've made an additional change to tidy up columns in the terminal output.

### DIFF
--- a/Sources/tart/Commands/List.swift
+++ b/Sources/tart/Commands/List.swift
@@ -7,7 +7,7 @@ struct List: AsyncParsableCommand {
 
   @Flag(name: [.short, .long], help: ArgumentHelp("Only display VM names."))
   var quiet: Bool = false
-  
+
   @Option(name: [.short, .long], help: ArgumentHelp("Only display VMs of this Source."))
   var source: String = ""
 
@@ -24,7 +24,7 @@ struct List: AsyncParsableCommand {
       default:
         throw ValidationError("Unknown source: '\(source)'")
       }
-      
+
       Foundation.exit(0)
     } catch {
       print(error)
@@ -37,7 +37,7 @@ struct List: AsyncParsableCommand {
     if !quiet {
       print("Source\tName")
     }
-    
+
     for (name, _) in vms.sorted(by: { left, right in left.0 < right.0 }) {
       if quiet {
         print(name)
--- a/Sources/tart/Commands/List.swift
+++ b/Sources/tart/Commands/List.swift
@@ -5,18 +5,30 @@ import SwiftUI
 struct List: AsyncParsableCommand {
   static var configuration = CommandConfiguration(abstract: "List created VMs")
 
-  @Flag(name: [.short, .long], help: ArgumentHelp("Only display VM names"))
+  @Flag(name: [.short, .long], help: ArgumentHelp("Only display VM names."))
   var quiet: Bool = false
+  
+  @Option(name: [.short, .long], help: ArgumentHelp("Only display VMs of this Source."))
+  var source: String = ""
 
   func run() async throws {
     do {
       if !quiet {
         print("Source\tName")
       }
-
-      displayTable("local", try VMStorageLocal().list())
-      displayTable("oci", try VMStorageOCI().list().map { (name, vmDir, _) in (name, vmDir) })
-
+      
+      switch source {
+      case "local":
+        displayTable("local", try VMStorageLocal().list())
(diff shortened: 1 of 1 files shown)


Sources/tart/Commands/List.swift
```diff
@@ -5,18 +5,26 @@ import SwiftUI
...
-      if !quiet {
-        print("Source\tName")
+      switch source {
+      case "local":
+        displayTable("local", try VMStorageLocal().list())
+      case "oci":
+        displayTable("oci", try VMStorageOCI().list().map { (name, vmDir, _) in (name, vmDir) })
+      case "":
+        displayTable("local", try VMStorageLocal().list())
+        displayTable("oci", try VMStorageOCI().list().map { (name, vmDir, _) in (name, vmDir) })
```
edigaryev, line 23: I think here we will print the header twice, because it is printed on each `displayTable()` call now.

You can probably move the validation logic into `validate()`, like in `tart delete`:

https://github.com/cirruslabs/tart/blob/7a2c20ba30e8e2e25a5b2ad035e2685cb9b54954/Sources/tart/Commands/Create.swift#L21-L25

...and revert the header change.

Sources/tart/Commands/List.swift
```diff
@@ -5,18 +5,26 @@ import SwiftUI
 struct List: AsyncParsableCommand {
   static var configuration = CommandConfiguration(abstract: "List created VMs")
 
-  @Flag(name: [.short, .long], help: ArgumentHelp("Only display VM names"))
+  @Flag(name: [.short, .long], help: ArgumentHelp("Only display VM names."))
   var quiet: Bool = false
 
+  @Option(name: [.short, .long], help: ArgumentHelp("Only display VMs of this Source."))
```
edigaryev, line 11: What do you think about:

* removing the short name for now, as we don't know if we may need `-s` in the future
* removing the capitalization for the word "source" and adding an example instead

```suggestion
  @Option(help: ArgumentHelp("Only display VMs from the specified source (e.g. --source local, --source oci)."))
```

Sources/tart/Commands/List.swift
marc-48k: @edigaryev if you think it makes more sense I could introduce an [EnumerableFlag](https://apple.github.io/swift-argument-parser/documentation/argumentparser/enumerableflag) so that the sources are validated automatically by ArgumentParser and mutual exclusivity is enforced.

Users would run `tart list --local` or `tart list --oci`. It depends what we have in mind in terms of sources for the future.

edigaryev: I don't really like this approach as it needlessly pollutes the flags namespace.

You have to manually search for all the possible variants (or name them like `--source-local`, `--source-oci`, etc., which makes little sense if we can just have `--source <source>`), instead of looking for a documentation for a single option, and it would be pain to deprecate them in the future.

I'm actually leaning more leaning towards Docker's approach, where there's only one `--filter` argument (see `docker image ls --help`, for example), but your current approach is sensible too, probably a good middle ground between `--filter=...` and `--local`/`--oci`.

marc-48k: Good points. 
I'll fix and move forward with my initial approach.

edigaryev (approved):

fkorotkov (approved):

fkorotkov:
Hey @marc-48k, 

We are preparing `1.0.0` release of Tart at some point in January and would like to clean things up and stabilize the API to provide long term support. Therefore we are introducing Contributor License Agreement which is common practice among mature OSS projects.

PR #351, opened 2022-12-14
Optimal buffer sizes for Softnet socket

fkorotkov:


fkorotkov:
@edigaryev how did you come up with the 1Mb buffer size? Do you think we need to change it as well?

edigaryev (approved):

edigaryev (approved):

PR #354, opened 2022-12-15
Report pull metrics to Sentry

fkorotkov:


Sources/tart/Root.swift
```diff
@@ -39,6 +39,14 @@ struct Root: AsyncParsableCommand {
       SentrySDK.start { options in
         options.dsn = dsn
         options.releaseName = CI.release
+        options.tracesSampleRate = 1.0
+
+        // By default only 5XX are captured
+        // Let's capture everything but 401 (unauthorized)
+        options.failedRequestStatusCodes = [
```
edigaryev, line 46: Should we also set `enableCaptureFailedRequests` for this to work?

fkorotkov: I think it's on by default but can certainly explicitly add it. 👌

Sources/tart/Commands/Prune.swift
```diff
@@ -103,5 +104,7 @@ struct Prune: AsyncParsableCommand {
     }
 
     puppy.info("reclaimed \(cacheReclaimedBytes) bytes")
+
+    SentrySDK.span?.setMeasurement(name: "gc_disk_reclaimed", value: cacheReclaimedBytes as NSNumber, unit: MeasurementUnitInformation.byte);
```
edigaryev, line 108: I think you need to [bind the transaction to scope](https://docs.sentry.io/platforms/apple/performance/instrumentation/custom-instrumentation/#create-transaction-bound-to-the-current-scope) the first, in order for the `SentrySDK.span` to work.

fkorotkov: My bad. I saw the documentation and though it's binded by default. 😪 Good catch!

edigaryev (dismissed):

edigaryev (dismissed):

PR #359, opened 2022-12-16
Don't throw if a VM is not running

fkorotkov:
Seems the exit code should be 0 similar to how we allow pull of a local image.

### DIFF
--- a/Sources/tart/Commands/Stop.swift
+++ b/Sources/tart/Commands/Stop.swift
@@ -19,7 +19,8 @@ struct Stop: AsyncParsableCommand {
     // Find the VM's PID
     var pid = try lock.pid()
     if pid == 0 {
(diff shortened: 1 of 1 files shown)


edigaryev:
I think this was done for a reason: https://github.com/cirruslabs/tart/pull/321.

But we could skip registering a Sentry error for exit codes other than 1, though. Somewhere around here:

https://github.com/cirruslabs/tart/blob/c3e37854c3f788ac2a3448387d8a5c87c01364a0/Sources/tart/Root.swift#L105-L107

What do you think?

fkorotkov:
We can ignore the issue on Sentry side as well but right now there is an issue that all `RuntimeError` are getting grouped together regardless of the message. We can probably fix that and then just ignore this "not running" error.

PR #364, opened 2022-12-20
support ZVirtioConsoleDevice for linux vm

peternguyen93:
Hi, I wanna contribute the controller for virtual machine via serial port.
User can now add this field into config.json to use serial port
```
{
  ...
  "serial": true
}
```
VM will create a serial device in **/dev/ttys00X**
user can use **screen /dev/ttys00X 115200** command to connect to the virtual machine.

### DIFF
--- a/Sources/tart/Serial.swift
+++ b/Sources/tart/Serial.swift
@@ -1,0 +1,43 @@
+//
+//  File.swift
+//  
+//
+//  Created by Peter Nguyen on 19/12/2022.
+//
+
+import Foundation
+
+func createPTY() -> Int32
+{
+    var tty_fd: Int32 = -1
+    var sfd: Int32 = -1
+    var termios_ = termios()
+    var tty_path = UnsafeMutablePointer<CChar>.allocate(capacity: 1024)
+    
+    let tty_fd_ref = UnsafeMutablePointer<Int32>.init(&tty_fd)
+    let sfd_ref = UnsafeMutablePointer<Int32>.init(&sfd)
+    let termios_ref = UnsafeMutablePointer<termios>.init(&termios_)
+    let unused_ref = UnsafeMutablePointer<winsize>.init(nil)
+    
+    var res = openpty(tty_fd_ref, sfd_ref, tty_path, termios_ref, unused_ref);
+    if(res < 0 ){
+        perror("openpty error")
+        return -1
+    }
+    
+    cfmakeraw(termios_ref)
+    if(tcsetattr(sfd, TCSAFLUSH, termios_ref) != 0){
+        perror("tcsetattr error")
+        return -1
+    }
+    
+    close(sfd)
+    
+    res = fcntl(tty_fd, F_GETFL)
+    fcntl(tty_fd, F_SETFL, res | O_NONBLOCK)
+    
+    print("Successfully open pty \(String(cString: tty_path))")
+    
+    tty_path.deallocate()
+    return tty_fd
+}
--- a/Sources/tart/Serial.swift
+++ b/Sources/tart/Serial.swift
@@ -14,18 +14,13 @@ func createPTY() -> Int32
   var termios_ = termios()
   let tty_path = UnsafeMutablePointer<CChar>.allocate(capacity: 1024)
     
-  var res = openpty(&tty_fd, &sfd, tty_path, &termios_, nil);
+  var res = openpty(&tty_fd, &sfd, tty_path, nil, nil);
   if(res < 0 ){
     perror("openpty error")
     return -1
   }
-        
-  cfmakeraw(&termios_)
-  if(tcsetattr(sfd, TCSANOW, &termios_) != 0){
-    perror("tcsetattr error")
-    return -1
-  }
     
+  // close slave file descriptor
   close(sfd)
         
   res = fcntl(tty_fd, F_GETFL)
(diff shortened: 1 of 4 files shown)


edigaryev:
Hi Peter. Thanks for your contribution!

Could you clarify a bit how are you planning to use this functionality?

Fox example, we could use [SwiftTerm](https://github.com/migueldeicaza/SwiftTerm) to automatically attach to a TTY (both in `tart run`'s terminal and in the Tart's UI) instead of manually asking the user to copy the TTY name and type `screen`.

peternguyen93:
hi @edigaryev I plan to use this feature to debug Linux kernel or macOS kernel via serial console which is more stable than debugging via udp network

edigaryev:
> hi @edigaryev I plan to use this feature to debug Linux kernel or macOS kernel via serial console which is more stable than debugging via udp network

Would displaying the TTY console Tart's UI work for you?

peternguyen93:
Hi @edigaryev,  Since macOS kernel debugging protocol works via serial port so I have to create a serial device to receive/send debug commands so I need to create a serial device for the virtual machine.

edigaryev:
> Hi @edigaryev, Since macOS kernel debugging protocol works via serial port so I have to create a serial device to receive/send debug commands so I need to create a serial device for the virtual machine.

The serial device will be created from the guest's side, but on the host's side we'll display a UI widget that will allow you to interact with the guest.

Both of these components will interact using [`Pipe`](https://developer.apple.com/documentation/foundation/pipe) and you won't need to `Ctrl+C`/`Ctrl-V` the TTY path each time you run the VM.

Sounds good?

peternguyen93:
Hi @edigaryev, I think add serial device for vm on host machine is quite useful for debugging Linux kernel/ macOS kernel since this feature doesn't work like copy/paste because it is protocol to communicate between debugger and KGDB or KDP.
The feature I describe above is just small case to use serial port on host machine, but I wanna use more features than just **screen /dev/ttyS00X**

edigaryev:
> but I wanna use more features than just **screen /dev/ttyS00X**

Do you want to use something like:

```
(gdb) target remote <tty-device>
```

?

peternguyen93:
hi @edigaryev that's right, I wanna do to attach to my debugger to debug Linux kernel and macOS kernel :))

Sources/tart/Serial.swift
```diff
@@ -0,0 +1,43 @@
...
+    var sfd: Int32 = -1
+    var termios_ = termios()
+    var tty_path = UnsafeMutablePointer<CChar>.allocate(capacity: 1024)
+    
+    let tty_fd_ref = UnsafeMutablePointer<Int32>.init(&tty_fd)
+    let sfd_ref = UnsafeMutablePointer<Int32>.init(&sfd)
+    let termios_ref = UnsafeMutablePointer<termios>.init(&termios_)
+    let unused_ref = UnsafeMutablePointer<winsize>.init(nil)
+    
+    var res = openpty(tty_fd_ref, sfd_ref, tty_path, termios_ref, unused_ref);
```
edigaryev, line 22: Two things here:

1. Should we pass `nil` instead of `termios_ref` here to avoid using zeroed `termios` structure for the initial pseudo-TTY configuration? Then, call `tcgetattr()` before `cfmakeraw()` to actually populate the `termios` structure before modifying it.

2. Can we just pass `nil` here instead of `unused_ref`, and do not declare `unused_ref` at all?

edigaryev:
Is there a reason why we need to have the serial port configuration the VM persistent? What do you think about just having a `--serial` (for example) argument to the `tart run` command?

Also, please fix the linter errors by running:

```shell
swift package plugin --allow-writing-to-package-directory swiftformat --cache ignore
```

Sources/tart/Serial.swift
peternguyen93: hi @edigaryev thank you for your reply I will fix these issues soon since I haven't familiar with swift before

peternguyen93: hi @edigaryev I have push some commits to clean code and add --serial option to tart

edigaryev: How about we remove serial port from the configuration then? See https://github.com/cirruslabs/tart/pull/364#issuecomment-1362813600.

edigaryev: Also, what do you think about item number 1 in https://github.com/cirruslabs/tart/pull/364#discussion_r1055435403?

peternguyen93: @edigaryev I agree with you to remove serial option in configuration file and I also removed it for the #1 I also cleared cfmakeraw because I didn't relate to tty setup.

dlon:
Any news on this?

faern:
Having this (and also available for macOS VMs) would be greatly appreciated by our project. We want to run automated tested in macOS VMs. And we can't have the test orchestration tools communicate with the VM over networking since the tests that run in the VM will mess a lot with the network stack of the guest. We are testing a VPN application. So we will add firewall rules, reconnect the network and do a bunch of other stuff that will not lend itself well to also having a network connection `host<->guest` at the same time.

We already have a solution for Linux + Windows VMs where we control the guests via virtual serial port. Would be awesome if we could do that with macOS guests as well.

fkorotkov (approved):

fkorotkov:
Merged and reformatted the code on top in [066f585](https://github.com/cirruslabs/tart/commit/066f585c534e5f15b28d3bfb66b790195bd848d2). On a side note in case of @faern's integration, it seems additionally will be useful to have an option to create a tty externally and just pass path to it. Kind of putting responsibility to the integrator so you don't need to parse output of Tart and have an option to customize PTY settings if needed.

What do you all think about creating a hidden `--serial-path` flag that you use to pass a tty directly?

faern:
Thank you so much for this! We currently have a suspicion that the code merged as part of this PR could also work for macOS guests for us. But we are evaluating that! But basically, what we need is any mechanism that allows us to get a raw IO mechanism between host and guest. We'll prepare our images so they start a system service on boot that connects to this IO and start accepting commands from the host on the outside. And it has to work no matter what the state of the network is in the guest.

fkorotkov:
@faern [1.1.0](https://github.com/cirruslabs/tart/releases/tag/1.1.0) just got release with this changes and #448. Please give it a try and let us know how it will go! Will be great to have Mullvad in the [list of users](https://github.com/cirruslabs/tart/blob/main/Resources/Users/HowToAddYourself.md).

Jontified:
We're using Tart as part of our testing framework for macOS. Currently everything seems to be working great, this was basically the one thing that was missing since, like @faern pointed out, we needed some form of communication channel between our VM test-manager and our host test-manager.
Thanks again!

PR #376, opened 2023-01-10
Clarify --dir requirements in README.md

chrisranderson:
Since Ventura is pretty new, I think it'd be good to warn people before they download a large image. `tart run --help` gives me this:

```
  --dir <name:path[:ro]>  Additional directory shares with an optional read-only specifier
                          (e.g. --dir="build:~/src/build" --dir="sources:~/src/sources:ro")
        Requires host to be macOS 13.0 (Ventura) or newer.
        All shared directories are automatically mounted to "/Volumes/My Shared Files" directory on macOS,
        while on Linux you have to do it manually: "mount -t virtiofs com.apple.virtio-fs.automount /mount/point".
        For macOS guests, they must be running macOS 13.0 (Ventura) or newer.
```

I'm unclear if `--dir` requires both host and guest to be running Ventura or if the "For macOS guests..." is only talking about the "All shared directories..." portion.

### DIFF
--- a/README.md
+++ b/README.md
@@ -278,7 +278,7 @@ This invocation calls the `tart pull` implicitly (if the image is not being pres
 
 ### Mounting directories
 
-**Note:** macOS 13.0 (Ventura) or newer is required on both the host and guest to mount a directory.
+**Note**: directory mounting is only supported on hosts running macOS 13.0 (Ventura) or newer. macOS guests must also use 13.0 or newer.
 
 To mount a directory, run the VM with the `--dir` argument:
 
--- a/README.md
+++ b/README.md
@@ -278,7 +278,7 @@ This invocation calls the `tart pull` implicitly (if the image is not being pres
 
 ### Mounting directories
 
-To mount a directory, run the VM with the `--dir` argument:
+To mount a directory, run the VM with the `--dir` argument (host must be macOS 13.0 Ventura or newer):
 
 ```shell
 tart run --dir=project:~/src/project vm
--- a/README.md
+++ b/README.md
@@ -278,7 +278,9 @@ This invocation calls the `tart pull` implicitly (if the image is not being pres
 
 ### Mounting directories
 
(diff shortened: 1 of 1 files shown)


chrisranderson:
If/when this is approved, I'll request CLA approval from my legal peeps.

edigaryev:
I think this is already mentioned in the `README.md` just a couple lines below:

>Note: to use the directory mounting feature, the host needs to run macOS 13.0 (Ventura) or newer.

chrisranderson:
It's a nitpick, but I think it's too low - you only need to read this portion to know what to do:

```
To mount a directory, run the VM with the --dir argument:

tart run --dir=project:~/src/project vm
```

The warning should come before usage, but it's currently several paragraphs below the example.

chrisranderson:
Is there a way to accomplish the same thing on < macOS 13.0 using `.cirrus.yml` and `cirrus run`? Documenting an alternative would be great.

edigaryev:
There's also a requirement for the guest to be macOS 13.0 in the "Accessing mounted directories in macOS guests" section:

>Note: to use the directory mounting feature, the guest VM needs to run macOS 13.0 (Ventura) or newer.

If we're removing the old note about host requirements and moving it upper in shorter form, I think the note about guest requirements needs to be dealt with too.

chrisranderson:
Fixed the note about guest requirements, adjusted it to match your style more closely, and put the OS at the beginning of the sentence to catch eyes.

I think it's okay only to put the warning there and remove the later note since you need to do that first before accessing the directory from the guest.

edigaryev:
Thanks! It looks nicer now.

However, the statement doesn't seem to be valid now for users that run Linux guests: they don't need macOS 13 (Ventura) guest to be able to mount directories.

Any ideas how this can be re-phrased to fix that?

edigaryev (approved):

fkorotkov (approved):

Issue #370, opened 2022-12-26
Import/Export without a remote OCI/ISO


tomjn:
Currently you can:

 - create a VM based off an ISO or IPSW
 - create a VM by pulling from an OCI registry

Neither of those are any good for vagrant VMs, so I've been copying the folder inside `~/.tart/vms` and changing the folder name to create a new VM. This technically works for creating vagrant boxes.

My concern though is that:

 - if the user can change where this VM folder is located
 - that the MAC address in the config JSON is identical
 - if the format of a VM folder ever changes it would invalidate all the old boxes

So it would be nice to be able to export locally the same information that gets pushed up to the OCI registry, as well as to import them. I imagine most of it would use the same code as the push/pull but without the network requests. This would also avoid the compatibility issue by reusing any solutions used for the OCI push/pull compatibility, and avoiding adding more code paths.

I also don't like messing with Tart internal folders if I can help it incase it breaks other functionality in Tart unrelated to what I'm attempting

PR #386, opened 2023-01-25
Introduce "tart import" and "tart export" commands

edigaryev:
Resolves https://github.com/cirruslabs/tart/issues/370.


fkorotkov (commented):
Have you looked into [something like this from Apple docs](https://developer.apple.com/documentation/accelerate/compressing_file_system_directories). Just arching the directory with the recommended algorithm. I wonder if there are some optimizations in the Apple Silicon chip itself.

edigaryev:
> Have you looked into [something like this from Apple docs](https://developer.apple.com/documentation/accelerate/compressing_file_system_directories). Just arching the directory with the recommended algorithm. I wonder if there are some optimizations in the Apple Silicon chip itself.

I've considered it, but this library has like 3 real-world usages [on GitHub](https://github.com/search?l=Swift&q=FieldKeySet&type=Code) and a very poor documentation. For example, [Define the Header Keys](https://developer.apple.com/documentation/accelerate/compressing_file_system_directories#3627227) guide step suggests the following cryptic field set:

```swift
guard let keySet = ArchiveHeader.FieldKeySet("TYP,PAT,LNK,DEV,DAT,UID,GID,MOD,FLG,MTM,BTM,CTM") else {
    return
}
```

While you might be able to find the explanation for the `TYP` and `PAT` somewhere in the documentation, the `LNK`, `DAT` and others doesn't seem to be documented.

Do you think the performance is that much critical for us here to keep looking into this library further?

fkorotkov:
First of all, I just feel that this change where you emulate a registry is very clever but unnecessary. Plus it seems we are compressing layers first and then compressing them once again in for the export. LZFSE is claimed to be optimized on M chips themselves. And just archiving a directory all together will be a smaller change and easier to read.

edigaryev:
@fkorotkov PTAL at https://github.com/cirruslabs/tart/pull/386/commits/04864adf86e6762613c9ad0ce7300b1e40668336, it addresses your concerns above.

Sources/tart/VMDirectory+Archive.swift
```diff
@@ -0,0 +1,90 @@
+import System
+import AppleArchive
+
+fileprivate let permissions = FilePermissions(rawValue: 0o644)
+
+extension VMDirectory {
```
fkorotkov, line 6: Could you please add a link to the Apple docs? And a note that `lzfse` is a recommended algorithm on Apple Platform.

fkorotkov (commented):
Looks awesome! Have you compared speed of this solution and the previous one?  Do you think we need to print at least something like `importing...`/`exporting...` since we can't get progress of the operation?

Sources/tart/VMDirectory+Archive.swift
edigaryev: See 46038002fd1bb1aaa4c201558f74d6305bb5ebd6.

edigaryev:
> Have you compared speed of this solution and the previous one?

I've just checked and it seems to be considerably faster, yes.

ZIPFoundation:

```
.build/arm64-apple-macosx/debug/tart export macos macos.zip  563.22s user 16.24s system 98% cpu 9:47.84 total
.build/arm64-apple-macosx/debug/tart export macos macos.zip  564.91s user 15.98s system 98% cpu 9:47.25 total
```

AppleArchive:

```
.build/arm64-apple-macosx/debug/tart export macos macos.aar  265.14s user 11.86s system 777% cpu 35.606 total
.build/arm64-apple-macosx/debug/tart export macos macos.aar  264.93s user 13.94s system 853% cpu 32.681 total
```

>Do you think we need to print at least something like `importing...`/`exporting...` since we can't get progress of the operation?

See 93d416e.

fkorotkov (approved):

PR #394, opened 2023-02-02
JSON output for get and list commands

### LINKED ISSUE #297: Machine friendly output
Tarts CLI commands output human friendly text, but for automation and integrations it's better to print in a machine readable format such as JSON or XML.

To this end the addition of a general `--json` parameter for all CLI commands would be a great help

fkorotkov:
In the light of the upcoming `1.0.0` release and stabilizing of the API, let's introduce some breaking changes for the good.

Removed all the `--cpu`, `--memory`, `--disk` and `--display` flags and replaced with a single `--json` flag for machine-readable output.

Added `--json` option to the `list` command to output a single JSON list. Notably removed `--quite` flag since it seemed unnecessary.

Fixes #297

### DIFF
--- a/Sources/tart/Commands/Get.swift
+++ b/Sources/tart/Commands/Get.swift
@@ -1,52 +1,29 @@
 import ArgumentParser
 import Foundation
 
+fileprivate struct VMInfo: Encodable {
+  let CPU: Int
+  let Memory: UInt64
+  let Disk: Int
+  let Display: String
+}
+
 struct Get: AsyncParsableCommand {
   static var configuration = CommandConfiguration(commandName: "get", abstract: "Get a VM's configuration")
 
   @Argument(help: "VM name.")
   var name: String
 
-  @Flag(help: "Number of VM CPUs.")
-  var cpu: Bool = false
-
-  @Flag(help: "VM memory size in megabytes.")
-  var memory: Bool = false
-
-  @Flag(help: "Disk size in gigabytes.")
-  var diskSize: Bool = false
-
-  @Flag(help: "VM display resolution in a format of <width>x<height>. For example, 1200x800.")
-  var display: Bool = false
-
-  func validate() throws {
-    if [cpu, memory, diskSize, display].filter({$0}).count > 1 {
-      throw ValidationError("Options --cpu, --memory, --disk-size and --display are mutually exclusive")
-    }
-  }
+  @Flag(help: "Output format")
+  var format: Format = .table
 
   func run() async throws {
     let vmDir = try VMStorageLocal().open(name)
     let vmConfig = try VMConfig(fromURL: vmDir.configURL)
     let diskSizeInGb = try vmDir.sizeBytes() / 1000 / 1000 / 1000
-    let memorySizeInMb = vmConfig.memorySize  / 1024 / 1024
+    let memorySizeInMb = vmConfig.memorySize / 1024 / 1024
 
-    if cpu {
-      print(vmConfig.cpuCount)
-    } else if memory {
-      print(memorySizeInMb)
-    } else if diskSize {
-      print(diskSizeInGb)
-    } else if display {
-      print("\(vmConfig.display.width)x\(vmConfig.display.height)")
-    } else {
-      print(
-        "CPU\tMemory\tDisk\tDisplay\n" +
-          "\(vmConfig.cpuCount)\t" +
-          "\(memorySizeInMb) MB\t" +
-          "\(diskSizeInGb) GB\t" +
-          "\(vmConfig.display)"
-      )
-    }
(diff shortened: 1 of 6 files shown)


fkorotkov:
FYI @tomjn @fmoc @hokadiri @josephburnett

edigaryev:
1. I do think it's beneficial to keep the `-q` and `--quiet` for simple use-cases like https://github.com/cirruslabs/tart/discussions/375#discussioncomment-4611062.

    I would never recommend anyone to parse table-like output that the `tart list` provides as this will likely break in the future (https://github.com/cirruslabs/tart/issues/297#issuecomment-1385814735), and processing `--json` output with `jq` might be too complicated for this every-day task.

2. IMO the VM size should be printed in human-understandable units and follow the source and name of the VM. `tart list` output when running this PR:

    ```
    Source Size Name                                                                                                         
    local  19   macos                                                                                                        
    oci    50   ghcr.io/cirruslabs/macos-ventura-base:latest
    ```

    Proposed `tart list` output:

    ````
    Source Name                                               Size    
    local  macos                                              18.45 GB
    oci    ghcr.io/cirruslabs/macos-ventura-base:latest       46.6 GB
    ````

4. The `--table/--json Output format (default: table)` option looks a bit weird, perhaps simply adding a `--json` flag should be enough?

5. There seems to be an extraneous newline emitted at the end when using `tart get` and `tart list`.

tomjn:
> and processing --json output with jq might be too complicated for this every-day task.

Without `jq` parsing it in shell script/bash becomes **significantly** harder, `jq` makes it much easier. If you're going to add JSON don't piecemeal it just add it everywhere in a consistent manner.

Use of JSON should also be an indicator that the intended recipient is inhuman

> IMO the VM size should be printed in human-understandable units and follow the source and name of the VM. 

This sounds like a great excuse for a flag/parameter and a new issue+PR

> The --table/--json Output format (default: table) option looks a bit weird, perhaps simply adding a --json flag should be enough?

Lets not forget YAML or real ASCII tables ( not pretend TSV's ), why not a `format` parameter?

WP CLI uses an optional `--format` parameter that works on all commands

```sh
wp post list # the default, same as table
wp post list --format=table
wp post list --format=json
wp post list --format=yaml
```

It also allows for `-q` and `--quiet`.

This is possible because there's a separation between doing things and generating output, with dedicated utilities just for formatting raw data into presentable text

https://make.wordpress.org/cli/handbook/references/internal-api/wp-cli-utils-format-items/

fmoc:
Docker for instance provides pretty-printed JSON when using `inspect`, which works well enough in the rare occasions you need to read those attributes manually. (With `jq`, you can even color-format the output, but in my opinion some default formatting/pretty-printing should be included so one does not depend on this tool.)

fkorotkov:
Thank you all for the feedback. I've addressed all of it except human readable disk size for the `text` output format. IMO for JSON is should be just an Integer value but I haven't found how to make formatting customization for `text` vs `json` output.

@edigaryev I've also reverted `--quite` flag in [f1dbe3e](https://github.com/cirruslabs/tart/pull/394/commits/f1dbe3ead60e485cc8b91fbe14afdd6d9677d444). Fixed new line in [b836429](https://github.com/cirruslabs/tart/pull/394/commits/b836429d5a8a1aa3df462613217913fed885f8b7). PTAL.

Sources/tart/Commands/Get.swift
```diff
@@ -1,52 +1,31 @@
...
-
-  @Flag(help: "VM display resolution in a format of <width>x<height>. For example, 1200x800.")
-  var display: Bool = false
-
-  func validate() throws {
-    if [cpu, memory, diskSize, display].filter({$0}).count > 1 {
-      throw ValidationError("Options --cpu, --memory, --disk-size and --display are mutually exclusive")
-    }
-  }
+  @Option(help: "Output format")
```
edigaryev, line 18:
```suggestion
@Option(help: "Output format: text or json")
```

Sources/tart/Commands/List.swift
```diff
@@ -2,15 +2,24 @@ import ArgumentParser
...
 struct List: AsyncParsableCommand {
   static var configuration = CommandConfiguration(abstract: "List created VMs")
 
-  @Flag(name: [.short, .long], help: ArgumentHelp("Only display VM names."))
-  var quiet: Bool = false
-
   @Option(help: ArgumentHelp("Only display VMs from the specified source (e.g. --source local, --source oci)."))
   var source: String?
 
+  @Option(help: "Output format")
```
edigaryev, line 17:
```suggestion
  @Option(help: "Output format: text or json")
```

edigaryev (approved):

PR #403, opened 2023-02-06
Add Krisp logo in users section

lanycrost:
Thank you for providing an excellent tool that has greatly enhanced our experience with Apple Silicon machines. We have utilized it to develop a small orchestration and scheduling system, which plays a crucial role in our CI/CD system.


fkorotkov (approved):
Loving Krisp! Great product and really glad you use Tart! 👌

edigaryev (approved):

PR #405, opened 2023-02-07
Ignore GC errors

fkorotkov:
Such errors are not critical


edigaryev:
Should we at least `catch` report these errors to `stderr`? 🤔

edigaryev (approved):

Issue #407, opened 2023-02-08
tart 0.37.1 release build test failure


chenrui333:
👋 trying to build the latest release, but run into some build issue. The error log is as below:

<details>
<summary>error test log</summary>

```
  ==> tart create --from-ipsw /private/tmp/tart-test-20230207-85543-1w88lax/empty.ipsw test 2>&1
  sh: line 1: 85558 Abort trap: 6           tart create --from-ipsw /private/tmp/tart-test-20230207-85543-1w88lax/empty.ipsw test 2>&1
  dyld[85558]: Symbol not found: (_$s10Foundation3URLV9formattedSSyF)
    Referenced from: '/opt/homebrew/Cellar/tart/0.37.1/bin/tart'
    Expected in: '/System/Library/Frameworks/Foundation.framework/Versions/C/Foundation'
```

</details>

full build log, https://github.com/Homebrew/homebrew-core/actions/runs/4117090987/jobs/7108059025
relates to Homebrew/homebrew-core#122584

PR #408, opened 2023-02-08
Avoid URL.formatted() method

edigaryev:
This method is not marked as macOS 13.0+ only in the SDK for some reason, but is marked so [in the documentation](https://developer.apple.com/documentation/foundation/url/4013498-formatted).

Resolves https://github.com/cirruslabs/tart/issues/407.


chenrui333:
Thanks for the patch, I am gonna try it now.

fkorotkov (approved):

PR #415, opened 2023-02-11
Relicensed under Fair Source License

fkorotkov:
As announced in https://tart.run/blog/2023/02/11/changing-tart-license/


LICENSE
```diff
@@ -1,661 +1,45 @@
...
 
-  When we speak of free software, we are referring to freedom, not
-price.  Our General Public Licenses are designed to make sure that you
-have the freedom to distribute copies of free software (and charge for
-them if you wish), that you receive source code or can get it if you
-want it, that you can change the software or use pieces of it in new
-free programs, and that you know you can do these things.
+License Grant. Licensor hereby grants to each recipient of the
+Software ("you") a non-exclusive, non-transferable, royalty-free and
+fully-paid-up license, under all of the Licensorâ€™s copyright and
```
edigaryev, line 14: >Licensorâ€™s

🤔

LICENSE
```diff
@@ -1,661 +1,45 @@
...
+patent rights, to use, copy, distribute, prepare derivative works of,
+publicly perform and display the Software, subject to the Use
+Limitation and the conditions set forth below.
 
-  Developers that use our General Public Licenses protect your rights
-with two steps: (1) assert copyright on the software, and (2) offer
-you this License which gives you legal permission to copy, distribute
-and/or modify the software.
+Use Limitation. The license granted above allows use by up to the
```
edigaryev, line 19: I think it'd be better to not use forced word-wrapping, since the original license text is not wrapped, so it would be easier for the users to diff the license text if they want to.

edigaryev (approved):

vitorgalvao:
I was unaware you had a blog or I’d have commented on this earlier. @fkorotkov Have you considered dual licensing for individuals/companies instead of changing the whole license outright?

Switching fully to a non-open-source license means tart will probably have to be [removed from Homebrew](https://docs.brew.sh/License-Guidelines) (and other package managers?), which then affects projects which depend on official formulae (i.e. not from third-party taps).

fkorotkov:
Hey @vitorgalvao, we considered duo licensing but in our opinion it would've brought additional complexity to already complex licensing question. Simplicity of Fair Software License is what attracted us in the first place, you don't need to have a degree in order to understand it.

Plus we made sure that Tart is continuing be loyalty free for vast majority of the users (note [the exception](https://github.com/cirruslabs/tart/blob/dbc1632fb4781b1de0abdaae19e6e379645829c1/LICENSE#L10) about devices that not shared aka your personal or work device and pretty large usage limitation on 100 cores).

As for the Homebrew, yes, `tart` should probably be removed from *homebrew-core* but you can always use [our tap](https://github.com/cirruslabs/homebrew-cli). In any case the `tart` binary built from sources from *homebrew-core* is not signed which makes some things not working:

- KeyChain access is wanky and requires `sudo` every time your upgrade the formula.
- Bridged Network doesn't work at all because it needs restricted entitlements.

vitorgalvao:
Thank you for the fast and thorough response. It’s unfortunate, but I understand your decision.

> the `tart` binary built from sources from *homebrew-core* is not signed

I see the one in [the releases](https://github.com/cirruslabs/tart/releases) is signed, but it is not notarised so it’s still killed when invoked on the command-line. You can’t directly notarise a binary, but you can zip it and notarise that. I gave a [short explanation of the process somewhere else](https://www.alfredforum.com/topic/19183-menu-bar-search/#comment-101234) but I can open a new issue for that if you prefer.

Unrelated, the blog doesn’t seem to have an RSS or JSON feed. Are those planned?

barkingfoodog:
+1 for an RSS blog feed. 

Could the .pkg be distributed as a [homebrew cask](https://github.com/Homebrew/homebrew-cask)?

vitorgalvao:
> Could the .pkg be distributed as a [homebrew cask](https://github.com/Homebrew/homebrew-cask)?

Yes, if the PKG is provided from an official place. Though that shouldn’t void the existence of a raw binary, which is more convenient in a lot of situations.

fkorotkov:
@vitorgalvao @barkingfoodog could you please create an issue about the RSS/JSON feed. Haven't seen such things so will appreciate if you can elaborate what you'd expect from such feeds.

fkorotkov:
Just released `1.0.1`. Binary inside the archive is notarized now as well.

PR #425, opened 2023-02-22
Hide `--net-bridged` flag

fkorotkov:
Until it's fixed in #243.


fkorotkov:
I actually figured out how to support bridged network in #429

PR #446, opened 2023-03-14
Allow to choose IP resolution strategy

fkorotkov:
Haven't seen any reports of the `arp` vs DHCP issues so I think it's reasonable to remove the warning and allow to customize the resolution strategy.

This way if one will want to access a Tart VM running on a remote host it will be possible:

```bash
remote-host:$ tart run --net-bridged=en0 my-vm
remote-host:$ tart ip --resolver=arp my-vm
192.168.192.222
local-host-on-the-same-network:$ ssh admin@192.168.192.222
```

### DIFF
--- a/Sources/tart/Commands/IP.swift
+++ b/Sources/tart/Commands/IP.swift
@@ -4,6 +4,12 @@ import Network
 import SystemConfiguration
 import Sentry
 
+enum IPResolutionStrategy: String, ExpressibleByArgument, CaseIterable {
+  case dhcp, arp
+
+  private(set) static var allValueStrings: [String] = Format.allCases.map { "\($0)"}
+}
+
 struct IP: AsyncParsableCommand {
   static var configuration = CommandConfiguration(abstract: "Get VM's IP address")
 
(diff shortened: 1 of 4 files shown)


Sources/tart/Commands/IP.swift
```diff
@@ -13,46 +19,40 @@ struct IP: AsyncParsableCommand {
   @Option(help: "Number of seconds to wait for a potential VM booting")
   var wait: UInt16 = 0
 
+  @Option(help: ArgumentHelp("Strategy for resolving IP adress: dhcp or arp",
+                             discussion: """
+                             By default, Tart is looking up and parsing DHCP lease file to determine aт IP for the VM.\n
+                             This method is fast and the most reliable but only returns local IP adresses.\n
+                             Alternatively, Tart can call external `arp` executable and parse it's output.\n
+                             In case of enabled Bridged Networking this method will return an IP adress on the network interface used for Bridged Networking.\n
```
edigaryev, line 27:
```suggestion
                             In case of enabled Bridged Networking this method will return VM's IP address on the network interface used for Bridged Networking.\n
```

Sources/tart/MACAddressResolver/Leases.swift
```diff
@@ -107,7 +107,7 @@ class Leases {
     return rawLeases
   }
 
-  func resolveMACAddress(macAddress: MACAddress) throws -> IPv4Address? {
+  func ResolveMACAddress(macAddress: MACAddress) throws -> IPv4Address? {
```
edigaryev, line 110: Why change the capitalization?

fkorotkov: So it's the same name as for the `arp`. I though about introducing a Protocol too but ended up not doing that. 🤷

edigaryev (approved):

PR #456, opened 2023-03-29
tart list: show if the VM is running or not

edigaryev:
See https://github.com/cirruslabs/orchard/pull/70#discussion_r1151842805.


fkorotkov (approved):
Oh! I thought we already report it. But it was only in `get` command.

fkorotkov (approved):

PR #459, opened 2023-03-30
UI improvements

### LINKED ISSUE #416: Tart UI: Reboot/restart/reload/shutdown
If a Tart VM is frozen and I have the GUI open, I have to close it, go back to the terminal, and re-run the command.

Instead it would be useful to have a restart/reload option in the top menu or a toolbar.

Likewise it would be useful to have an explicit shutdown option

### LINKED ISSUE #417: UI: VM info
There's a Tart about screen, and the VM name is in the title, but it would be useful to have the contents of `tart get` in a quick read only info box too

### LINKED ISSUE #418: Tart UI: Take a Screenshot
It's difficult to take screenshots of VMs as the common keyboard shortcuts are sent to the guest VM not the host.

The experimental VNC option fires up the screen sharing app by Apple which has a screenshot menu option, a similar option in Tart would be helpful

fkorotkov:
Allow to start/stop VM from a menu item of the Window. Fixes #416
Don't capture system keys so it's possible to take screenshots. Fixes #418 
Included VM info in the "About" dialog. Fixes #417


fkorotkov:
Also noticed that AppCode started building it's binaries in it's own derived data directory and therefore our pre-defined run configuration stopped working. Added a simple script that builds and signs binaries until Swift Package Manager can do the same.

Sources/tart/Commands/Run.swift
```diff
@@ -424,14 +438,42 @@ struct Run: AsyncParsableCommand {
...
+  var credits: NSAttributedString
+
+  init(config: VMConfig) {
+    let mutableAttrStr = NSMutableAttributedString()
+    let style = NSMutableParagraphStyle()
+    style.alignment = NSTextAlignment.center
+    let attrCenter: [NSAttributedString.Key : Any] = [
+      .paragraphStyle: style,
+    ]
+    mutableAttrStr.append(NSAttributedString(string: "CPU: \(config.cpuCount)\n", attributes: attrCenter))
```
edigaryev, line 460:
```suggestion
    mutableAttrStr.append(NSAttributedString(string: "CPU: \(config.cpuCount) cores\n", attributes: attrCenter))
```

Sources/tart/Commands/Run.swift
```diff
@@ -424,14 +438,42 @@ struct Run: AsyncParsableCommand {
...
+    mutableAttrStr.append(NSAttributedString(string: "Memory: \(config.memorySize / 1024 / 1024)\n", attributes: attrCenter))
```
edigaryev, line 461:
```suggestion
    mutableAttrStr.append(NSAttributedString(string: "Memory: \(config.memorySize / 1024 / 1024) MB\n", attributes: attrCenter))
```

edigaryev (approved):

Issue #423, opened 2023-02-21
`tart tag` command


fkorotkov:
So it's easier to create a new tag from an existing tag without using Docker or pushing the same image under a new tag:

```bash
tart tag <remote image>:existing-tag <remote image>:new-tag
```

PR #461, opened 2023-04-03
Introduce "tart tag" command to tag OCI VM images

edigaryev:
Resolves https://github.com/cirruslabs/tart/issues/423.

### DIFF
--- a/Sources/tart/Commands/Tag.swift
+++ b/Sources/tart/Commands/Tag.swift
@@ -1,0 +1,34 @@
+import ArgumentParser
+import Foundation
+
+struct Tag: AsyncParsableCommand {
+  static var configuration = CommandConfiguration(commandName: "tag", abstract: "Tag an OCI VM",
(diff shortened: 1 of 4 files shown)


edigaryev:
> I think `tag` command should also push a new tag with the same sha256 to the remote registry before "fixing" local storage.

I think this will be greatly misleading for people coming from Docker, as it's `docker tag` command doesn't do any pushing.

Re-reading the https://github.com/cirruslabs/tart/issues/423 again, I think the proper way to achieve "pushing the same image under a new tag" without causing confusion would be something like this:

1. Allow to pass an OCI VM in the first argument of `tart push`.
2. Use the manifest digest (`sha256:...`) of the local OCI VM to avoid needless compression and digest computation in case the remote blobs already exist.

What do you think?

edigaryev:
Closing in favor of https://github.com/cirruslabs/tart/pull/465.

PR #464, opened 2023-04-05
Improve `tart ip` error message

### LINKED ISSUE #460: `tart ip` returns invalid IP when running a Linux VM with `--net-bridged`
Running VMs with the "net-bridged" adapter is excellent, but the `tart ip`  still prints the NAT address of the VM:
```
% tart run --rosetta rosetta --net-bridged en0 myvm-1234 
% tart ip myvm-1234
192.168.64.6
```
Using `--resolver arp` returns `no IP address found`:
```
% tart ip --resolver arp myvm-1234
no IP address found, is your VM running?
% tart ip --resolver=arp myvm-1234
no IP address found, is your VM running?
```

This issue is important because the packer plugin uses the same command ([here](https://github.com/cirruslabs/packer-plugin-tart/blob/main/builder/tart/ssh.go#L9)) to retrieve the IP address and initiate an SSH session while the builder creates artefacts. 
Regards

fkorotkov:
Resolves #460

### DIFF
--- a/Sources/tart/Commands/IP.swift
+++ b/Sources/tart/Commands/IP.swift
@@ -38,7 +38,7 @@ struct IP: AsyncParsableCommand {
       throw RuntimeError.NoIPAddressFound("no IP address found, is your VM running?")
     }
 
-    guard let ip =  ip = try await IP.resolveIP(vmMACAddress, resolutionStrategy: resolver, secondsToWait: wait) else {
+    guard let ip = try await IP.resolveIP(vmMACAddress, resolutionStrategy: resolver, secondsToWait: wait) else {
       if (vmConfig.os == .linux && resolver == .arp) {
         throw RuntimeError.NoIPAddressFound("no IP address found, not all Linux distributions are compatible with ARP resolver")
       }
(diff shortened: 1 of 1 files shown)


Sources/tart/Commands/IP.swift
```diff
@@ -34,10 +34,20 @@ struct IP: AsyncParsableCommand {
     let vmConfig = try VMConfig.init(fromURL: vmDir.configURL)
     let vmMACAddress = MACAddress(fromString: vmConfig.macAddress.string)!
 
-    guard let ip = try await IP.resolveIP(vmMACAddress, resolutionStrategy: resolver, secondsToWait: wait) else {
+    let lock = try PIDLock(lockURL: vmDir.configURL)
+    if try lock.pid() == 0 {
```
edigaryev, line 38: I think you can simplify this by using `vmDir.running()` instead of instantiating the `PIDLock` here.

Sources/tart/Commands/IP.swift
```diff
@@ -34,10 +34,20 @@ struct IP: AsyncParsableCommand {
...
       throw RuntimeError.NoIPAddressFound("no IP address found, is your VM running?")
     }
-    print(ip)
+
+    let ip = try await IP.resolveIP(vmMACAddress, resolutionStrategy: resolver, secondsToWait: wait)
```
edigaryev, line 42: Is there a reason to stop using `guard` here (as it was before)?

You can avoid things like `ip == nil` and `ip!` below if you'll keep using it.

Sources/tart/Commands/IP.swift
fkorotkov: Oh! Forgot we extracted the logic in a method. 👌 Fixed in 2487a844f5b839d9d51d4564c092c87a5f865ea8

edigaryev (approved):

PR #466, opened 2023-04-06
Refactored pushing of OCI images under new tag

fkorotkov:
Simplification that I implied in https://github.com/cirruslabs/tart/pull/465#discussion_r1159038157

### DIFF
--- a/Sources/tart/Commands/Push.swift
+++ b/Sources/tart/Commands/Push.swift
@@ -28,6 +28,7 @@ struct Push: AsyncParsableCommand {
   var populateCache: Bool = false
 
   func run() async throws {
+    let ociStorage = VMStorageOCI()
     let localVMDir = try VMStorageHelper.open(localName)
 
     // Parse remote names supplied by the user
--- a/Sources/tart/Commands/Push.swift
+++ b/Sources/tart/Commands/Push.swift
@@ -6,7 +6,7 @@ import Compression
 struct Push: AsyncParsableCommand {
   static var configuration = CommandConfiguration(abstract: "Push a VM to a registry")
 
-  @Argument(help: "local VM name")
+  @Argument(help: "local or remote VM name")
   var localName: String
 
   @Argument(help: "remote VM name(s)")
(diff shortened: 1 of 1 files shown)


Sources/tart/Commands/Push.swift
```diff
@@ -54,48 +54,43 @@ struct Push: AsyncParsableCommand {
...
+      let pushedRemoteName: RemoteName
       // If we're pushing a local OCI VM, check if points to an already existing registry manifest
       // and if so, only upload manifests (without config, disk and NVRAM) to the user-specified references
-      if let remoteName = try? RemoteName(localName),
-         let remoteName = try await lightweightPushToRegistry(registry: registry, remoteName: remoteName,
-                                                              references: references) {
-        pushedRemoteName = remoteName
-        lightweightPush = true
+      if let remoteName = try? RemoteName(localName) {
+        pushedRemoteName = try await lightweightPushToRegistry(
```
edigaryev, line 62: Imagine we have an OCI image that is not present in the registry. Because we don't have an additional condition as we have in https://github.com/cirruslabs/tart/pull/465 before jumping into this branch, `tart push` will be of no use for uploading that OCI image to the registry.

Sources/tart/Commands/Push.swift
```diff
@@ -54,48 +54,43 @@ struct Push: AsyncParsableCommand {
...
 
-  func lightweightPushToRegistry(registry: Registry, remoteName: RemoteName, references: [String]) async throws -> RemoteName? {
+  func lightweightPushToRegistry(registry: Registry, remoteName: RemoteName, references: [String]) async throws -> RemoteName {
     // Is the local OCI VM already present in the registry?
     let digest = try VMStorageOCI().digest(remoteName)
 
-    guard let (remoteManifest, _) = try? await registry.pullManifest(reference: digest) else {
-      return nil
-    }
+    let (remoteManifest, _) = try await registry.pullManifest(reference: digest)
```
edigaryev, line 93: This will throw an exception if the remote reference does not exist.

This is not desirable, because we want to be opportunistic here: try a lightweight path first and then fallback to a heavyweight one if the former fails.

Sources/tart/Commands/Push.swift
fkorotkov: I think it's a reasonable assumption. If you have an OCI image locally, you pulled it from somewhere. If it's missing remotely, the layers might miss as well and it's reasonable to fail here IMO.

Sources/tart/Commands/Push.swift
fkorotkov: See my previous comment. I think it's reasonable to fail here.

Sources/tart/Commands/Push.swift
edigaryev: >If it's missing remotely, the layers might miss as well and it's reasonable to fail here IMO.

Sorry, but I don't think it is reasonable at all: your proposal essentially prevents the users from uploading these layers again to the registry in the event of their deletion on the registry's side.

edigaryev (approved):

PR #490, opened 2023-05-06
Pass through env variables to credential helpers

fkorotkov:
Fixes #488

### DIFF
--- a/Sources/tart/Credentials/DockerConfigCredentialsProvider.swift
+++ b/Sources/tart/Credentials/DockerConfigCredentialsProvider.swift
@@ -25,6 +25,8 @@ class DockerConfigCredentialsProvider: CredentialsProvider {
 
     let process = Process.init()
     process.executableURL = executableURL
+    // pass environment variables so things like AWS_PROFILE are picked up
+    process.environment = ProcessInfo.processInfo.environment
     process.arguments = ["get"]
 
     let outPipe = Pipe()


Sources/tart/Credentials/DockerConfigCredentialsProvider.swift
```diff
@@ -25,6 +25,8 @@ class DockerConfigCredentialsProvider: CredentialsProvider {
 
     let process = Process.init()
     process.executableURL = executableURL
+    // pass environment variables so things like AWS_PROFILE are picked up
+    process.environment = ProcessInfo.processInfo.environment
```
edigaryev, line 29: I think this is a no-op:

>By default, a Process object inherits its environment from the process that launches it.

(from https://developer.apple.com/documentation/foundation/process)

edigaryev:
Judging from the `~/.docker/config.json` in https://github.com/cirruslabs/tart/discussions/488, I think the issue is that we just don't support `credsStore` at all, [only `auths` and `credHelpers`](https://github.com/cirruslabs/tart/blob/ad9c3c661ea467e3e4caaef6f6f468c06528a503/Sources/tart/Credentials/DockerConfigCredentialsProvider.swift#L59-L62).

Sources/tart/Credentials/DockerConfigCredentialsProvider.swift
fkorotkov: My bad. Seems user should just switch to newer `credHelpers`.

Issue #479, opened 2023-04-27
VMs do not get assigned IPs regularly


fmoc:
It seems to be a known problem that when cloning VMs, one may not get an IP address assigned. Therefore (and because I often need to make minor adjustments) I create almost all VMs using packer, which works well most of the time.

However, sometimes (and I cannot tell exactly when or why), the VMs do not get assigned an IP address. This cannot be detected unless the output of `tart ip ...` is string-matched against a known value. There is no way to fix this (although at first, rebooting seemed to help, it does not). The VM must be deleted and recreated, otherwise there is no chance of receiving an IP address.

As this is an intermittent problem, I cannot provide steps how to reliably reproduce this bug. I can tell from experience that bug rarely occurs on two hosts which run a single tart VM created by configuration management, and happens regularly on another host which hosts two VMs also created by configuration management (usually in sequence with little delay).

Please let me know what kind of information is useful to you to debug this problem.

fkorotkov:
Hey @fmoc, could you please provide some details of how the VMs are getting created? Do you do a lot of `clone/run/delete` loop for running some tests? If so, do you clone from a local VM or a remote VM that is cached?

fmoc:
That was causing it initially, yes. Now, I `tart delete` and `packer build` only on changes, maybe once a date, since the tart VMs are used as CI runners now. Packer clones a local VM.

fkorotkov:
When Tart clones a VM it checks that all local VMs have unique MAC addresses and generates a new one if needed. In your case Tart will always generate a new MAC address which might contribute to getting the flake. Ideally, for `clone/run/delete` loop we recommend to clone from a remote VM specifically to make sure everything is more stable. We follow this principle for Cirrus CI and Cirrus Runners where we have dozens and sometimes 100+ such loops every day without similar issues.

BTW I assumed that you have a macOS VM. For Linux VMs situation is slightly different and depends on the distribution.

edigaryev:
Do you create more than 253 new VMs per day?

If so, this might be a [DHCP exhaustion](https://github.com/cirruslabs/softnet#working-model) and can be solved by re-using MAC-addresses or using [Softnet](https://github.com/cirruslabs/softnet), which has a fix that decreases the DHCP lease time from 1 day to 10 minutes.

fmoc:
[...]

PR #494, opened 2023-05-09
FAQ: document how to change the default DHCP lease time

edigaryev:
So that using Softnet will not the only way to fix DHCP exhaustion problem.

See https://github.com/cirruslabs/tart/issues/479.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -54,3 +54,13 @@ The default subnet mask `255.255.255.0` should suffice for most use-cases, howev
   ```bash
   sudo defaults write /Library/Preferences/SystemConfiguration/com.apple.vmnet.plist Shared_Net_Mask -string 255.255.0.0
   ```
+
+## Changing the default DHCP lease time
+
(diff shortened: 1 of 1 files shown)


docs/faq.md
```diff
@@ -43,14 +43,26 @@ is stricter and it's not only possible to access the host.
...
+```
+
+## Changing the default DHCP lease time
+
+By default, the built-in macOS DHCP server allocates IP-addresses to the VMs for the duration of 86,400 seconds (one day), which may easily cause DHCP exhaustion if you run more than ~253 VMs per day, or in other words, more than one VM every ~6 minutes.
+
+This issue is worked around automatically [when using Softnet](http://github.com/cirruslabs/softnet), however, if you don't use or can't use it, the following command will reduce the lease time from the default 86,400 seconds (one day) to 600 seconds (10 minutes):
+
+```shell
+sudo defaults write /Library/Preferences/SystemConfiguration/com.apple.InternetSharing.default.plist bootpd -dict DHCPLeaseTimeSecs -int 600
```
fkorotkov, line 65: Do you need to reboot after it?

edigaryev: > Do you need to reboot after it?

Nope.

fkorotkov (approved):

PR #506, opened 2023-05-26
Clarify Licensing Use Limitation

fkorotkov:
For both Tart and Orchard

[skip ci]

### DIFF
--- a/docs/licensing.md
+++ b/docs/licensing.md
@@ -4,8 +4,9 @@ hide:
 ---
 
 Both [Tart Virtualization](https://github.com/cirruslabs/tart) and [Orchard Orchestration](https://github.com/cirruslabs/orchard)
-are licensed under [Fair Source 100 License](https://fair.io/). Usage on personal computers including personal workstations is royalty-free,
-but organizations that exceed a certain number of server installations utilizing 100 CPU cores will be required to obtain a paid sponsorship.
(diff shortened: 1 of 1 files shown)


docs/licensing.md
```diff
@@ -4,8 +4,9 @@ hide:
 ---
 
 Both [Tart Virtualization](https://github.com/cirruslabs/tart) and [Orchard Orchestration](https://github.com/cirruslabs/orchard)
-are licensed under [Fair Source 100 License](https://fair.io/). Usage on personal computers including personal workstations is royalty-free,
-but organizations that exceed a certain number of server installations utilizing 100 CPU cores will be required to obtain a paid sponsorship.
+are licensed under [Fair Source License](https://fair.io/). Usage on personal computers including personal workstations is royalty-free,
+but organizations that exceed a certain number of server installations (100 CPU cores for Tart and/or 4 hosts for Orchard)
```
edigaryev, line 8: LGTM overall, but are you sure that we need to mention Orchard here?

Orchard [has it's own license](https://github.com/cirruslabs/orchard/blob/main/LICENSE), and we can change it the same way separately, also this is the only mention of Orchard in this legal text, and to some it might not be evident what we're talking about here.

fkorotkov: Yeah, but the sponsorship agreement covers both licenses. I think it's fair to have it here until Orchard is that big we have a separate site.

edigaryev (approved):

PR #510, opened 2023-06-02
Replace mentions of Monteray with Ventura in quickstart guide

fiedukow:
They seem to just be leftovers from previous versions of the docs, so a small cleanup could be helpful.

### DIFF
--- a/docs/quick-start.md
+++ b/docs/quick-start.md
@@ -3,7 +3,7 @@ hide:
   - navigation
 ---
 
-Try running a Tart VM on your Apple Silicon device running macOS 12.0 (Ventura) or later (will download a 25 GB image):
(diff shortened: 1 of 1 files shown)


docs/quick-start.md
```diff
@@ -3,7 +3,7 @@ hide:
   - navigation
 ---
 
-Try running a Tart VM on your Apple Silicon device running macOS 12.0 (Monterey) or later (will download a 25 GB image):
+Try running a Tart VM on your Apple Silicon device running macOS 12.0 (Ventura) or later (will download a 25 GB image):
```
edigaryev, line 6: Perhaps this should be left as is since Tart still supports macOS 12.0 (Monterey) as a host.

fiedukow: Oh, right, reverted that.

edigaryev (approved):

fkorotkov (approved):

PR #524, opened 2023-06-23
Use Mac-specific input devices when possible

edigaryev:


Sources/tart/Platform/Darwin.swift
```diff
@@ -98,11 +98,17 @@ struct Darwin: Platform {
...
+    } else {
+      return [VZUSBKeyboardConfiguration()]
+    }
+  }
+
   func pointingDevices() -> [VZPointingDeviceConfiguration] {
     if #available(macOS 13, *) {
-      // Trackpad is only supported starting with macOS Ventura
-      // macOS Monterey will continue using a USB device == .darwin 
-      return [VZMacTrackpadConfiguration(), VZUSBScreenCoordinatePointingDeviceConfiguration()]
```
fkorotkov, line 105: Will it make Monterey guests not having any pointing devices?

edigaryev: As far as I understood from [`VZMacTrackpadConfiguration` documentation](https://developer.apple.com/documentation/virtualization/vzmactrackpadconfiguration), the underlying device is selected automatically:

>In macOS 13 and later, guests use the multi-touch trackpad device, while earlier versions of macOS uses the USB pointing device.

I've tried running this change with a Monterey guest, and it works.

fkorotkov (approved):

Issue #147, opened 2022-07-12
Live snapshotting in tart


jerrymarino:
Good morning!

I was looking to do some "live" snapshotting macOS VMs with tart, where memory and state is suspended and then continuously re-serialized 

Mainly I'd do something like this:
```
tart run foo
.. open some programs, write some memory
tart clone foo bar
```

When running bar, I'd like to have the programs running resume from where they last were at: e.g. counters, ram, etc
```
tart run bar
```

Is this one on your radar or a reasonable feature to upstream?

fkorotkov:
There is this [`pause` method](https://developer.apple.com/documentation/virtualization/vzvirtualmachine/3656824-pause) that we haven't investigated yet. The documentation is pretty vague and I'm not sure how it can preserve RAM (haven't seen any reference to anything like this). I know in Anka version 2 suspending itself is pretty challenging since your VM was launched and bunch of stuff like networking was already initialized so resuming of such suspended VM is not trivial.

If that `pause` method does what you expect then I think we can add `Puase` button to the UI. Otherwise we'll need to wait for some kind of support from the `Virtualizaion.Framework`.

edigaryev:
Will keep this open as it seems that other `not possible atm` issues are open too.

jerrymarino:
Cool thanks for the heads up on this. Yeah it seems like it'd probably need some finessing to say the least 🙃 . For context, I have a number of features that I'd like to add to tart or file an issue for. I am attempting to use tart this to spin up throwaway test environments; booting up macOS to run a single test and throwing it away. I might have 1-2 other ones to file around this use case

> Regarding the [pause method](https://developer.apple.com/documentation/virtualization/vzvirtualmachine/3656824-pause), it doesn't seem to cover the use-case described as it only stops the vCPU from executing more instructions.

Since this method is on a public interface for an Apple VM system, maybe it'd be prudent on my end to feedbacks to ensure it works. I will also give this a a try too. 

With some of the open source hypervisor / VM tech out there, this works to some extent but with caveats: e.g. disconnecting virtio devices, and taking other steps to making sure it'd handle the resume

edigaryev:
[...]

PR #527, opened 2023-06-27
Suspend/resume support

edigaryev:
Unfortunately, changing the MAC-address after the snapshot is created is not possible (`Virtualization.Framework` throws an invalid argument exception when restoring), so we're not preserving the VM state when `tart clone`'ing.

Resolves https://github.com/cirruslabs/tart/issues/147.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -6,6 +6,7 @@ env:
 task:
   name: Test on Ventura
   alias: test
+  use_compute_credits: $CIRRUS_USER_COLLABORATOR == 'true'
   persistent_worker:
     labels:
       name: dev-mini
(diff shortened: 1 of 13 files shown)


fkorotkov (commented):
What happens when an audio and entropy devices are attached? Is it crashing. IMO it should work with them. Weird that it's not working and maybe newer versions of Sonoma will fix it.

Also what do you think of making VM to suspend on the current `Cmd + C` or closing the window?

edigaryev:
> Is it crashing.

Yes, see https://github.com/cirruslabs/tart/issues/147#issuecomment-1580969378.

edigaryev:
> Also what do you think of making VM to suspend on the current `Cmd + C` or closing the window?

How would the user shutdown the VM without taking a snapshot then?

Doing a clean "Shut Down" from inside of a VM is slower than simply closing the window and might not always be an option when the VM is unresponsive.

edigaryev:
Enabled automatic suspending of the VM when closing Tart's window and using `--suspendable`.

We can think of better ways of detecting whether we need to suspend or stop the VM later, when `--suspendable` is not needed anymore.

fkorotkov (approved):

PR #531, opened 2023-06-28
Optimistically check if we need to do anything on a pull

fkorotkov:
Right now on a pull we always acquire a lock for a registry host. This is problematic. For example, a worker can be pulling `ghcr.io/cirruslabs/macos-ventura-xcode:15-beta-2` image when a new request will come to pull `ghcr.io/cirruslabs/macos-ventura-xcode:latest` if needed.

In this situation, even though `ghcr.io/cirruslabs/macos-ventura-xcode:latest` is already cached and linked, `tart pull` will wait for a lock because the lock is per registry host.

This change optimistically check if there is something to do at all before acquiring a lock.

### DIFF
--- a/Sources/tart/VMStorageOCI.swift
+++ b/Sources/tart/VMStorageOCI.swift
@@ -139,6 +139,12 @@ class VMStorageOCI: PrunableStorage {
 
     let digestName = RemoteName(host: name.host, namespace: name.namespace,
                                 reference: Reference(digest: Digest.hash(manifestData)))
+    
+    if exists(name) && exists(digestName) && linked(from: name, to: digestName) {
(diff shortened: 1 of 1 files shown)


edigaryev (dismissed):

edigaryev (approved):

edigaryev (changes requested):
What do you think about moving the already existing locking functionality into the `if !exists(digestName) {` code block instead?

This way we wouldn't need to introduce duplicate "already cached" checks.

fkorotkov:
I was thinking about it but in the end there is the `link` method which is not atomic. It deletes a symlink and then creates new in to separate calls. Theoretically there can be a race on creation.

edigaryev (approved):

PR #532, opened 2023-06-29
tart clone: try to reclaim disk space if needed

edigaryev:
To avoid cases when OCI images take large space amounts and prevent the new VMs from being cloned.


fkorotkov (approved):

fkorotkov (approved):
I was checking if `SentrySDK.configureScope` is setting things up globally and it does. But I guess it's fine for a single invocation thing like Tart. In Cirrus CLI we do things differently and pass things in callbacks.

PR #539, opened 2023-07-04
Snapshotting Improvements

fkorotkov:
I think in order to satisfy users' use case of using a suspended VM with Tartlet we need to name MAC address generation lazy. E.g. check if there is already a running VM with the same MAC only when we invoke `tart run` instead of generating it on every `tart clone`. This way we can also clone suspended state of a VM so one can have a locally suspended "golden" VM which will be used for `clon`/`run`/`delete` activity of Tartlet, Cirrus CLI or any other integration.

Plus fixed some bugs from the original PR. Please see first few commits.

### DIFF
--- a/Sources/tart/Commands/Push.swift
+++ b/Sources/tart/Commands/Push.swift
@@ -74,7 +74,9 @@ struct Push: AsyncParsableCommand {
         // Populate the local cache (if requested)
         if populateCache {
           let expectedPushedVMDir = try ociStorage.create(pushedRemoteName)
-          try localVMDir.clone(to: expectedPushedVMDir)
+          // there might be a case when two remote VMs have identical MAC addresses
+          let generateMAC = try VMStorageLocal().hasVMsWithMACAddress(macAddress: expectedPushedVMDir.macAddress())
+          try localVMDir.clone(to: expectedPushedVMDir, generateMAC: generateMAC)
         }
       }
 
--- a/Sources/tart/Commands/Push.swift
+++ b/Sources/tart/Commands/Push.swift
@@ -74,9 +74,7 @@ struct Push: AsyncParsableCommand {
         // Populate the local cache (if requested)
         if populateCache {
           let expectedPushedVMDir = try ociStorage.create(pushedRemoteName)
-          // there might be a case when two remote VMs have identical MAC addresses
-          let generateMAC = try VMStorageLocal().hasVMsWithMACAddress(macAddress: expectedPushedVMDir.macAddress())
-          try localVMDir.clone(to: expectedPushedVMDir, generateMAC: generateMAC)
+          try localVMDir.clone(to: expectedPushedVMDir, generateMAC: false)
         }
       }
 
--- a/Sources/tart/Commands/Push.swift
+++ b/Sources/tart/Commands/Push.swift
@@ -74,7 +74,7 @@ struct Push: AsyncParsableCommand {
         // Populate the local cache (if requested)
         if populateCache {
           let expectedPushedVMDir = try ociStorage.create(pushedRemoteName)
-          try localVMDir.clone(to: expectedPushedVMDir, generateMAC: false)
+          try localVMDir.clone(to: expectedPushedVMDir)
         }
       }
 
--- a/Sources/tart/Commands/Run.swift
+++ b/Sources/tart/Commands/Run.swift
@@ -112,7 +112,33 @@ struct Run: AsyncParsableCommand {
 
   @MainActor
   func run() async throws {
-    let vmDir = try VMStorageLocal().open(name)
+    let localStorage = VMStorageLocal()
+    let vmDir = try localStorage.open(name)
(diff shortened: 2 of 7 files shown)


Sources/tart/Commands/Run.swift
```diff
@@ -239,10 +251,15 @@ struct Run: AsyncParsableCommand {
     }
     sigintSrc.activate()
 
-    // "tart suspend" support
+    // "tart suspend" / UI window closing support
     signal(SIGUSR1, SIG_IGN)
     let sigusr1Src = DispatchSource.makeSignalSource(signal: SIGUSR1)
     sigusr1Src.setEventHandler {
+      if !suspendable {
```
edigaryev, line 258: With this approach the caller of `tart run` won't be able to detect an error when using `tart suspend` and they'll assume that the VM was suspended successfully.

Can we pass the `suspendable` to the `MainApp` instead and use the old logic of choosing either `SIGUSR1`/`SIGINT` as a `kill(2)` argument?

fkorotkov: Not sure how easily pass the `suspendable` variable into `MainApp`. Let me investigate.

fkorotkov: PTAL [020e2d4](https://github.com/cirruslabs/tart/pull/539/commits/020e2d469e68c619f584476ef2a19e0f2b1f7e0f)

Sources/tart/Commands/Run.swift
```diff
@@ -112,7 +112,19 @@ struct Run: AsyncParsableCommand {
...
   func run() async throws {
-    let vmDir = try VMStorageLocal().open(name)
+    let localStorage = VMStorageLocal()
+    let vmDir = try localStorage.open(name)
+    let needToGenerateNewMac = try localStorage.list().contains {
+      // check if there is a running VM with the same MAC but different name
+      try $1.running() && $1.macAddress() == vmDir.macAddress() && $1.name != vmDir.name
+    }
+
+    // generate new MAC only after locking to make sure it's not this VM which is running
```
edigaryev, line 122: The comment seems to be misleading: there's no locking done here.

We should probably at least take a `TART_HOME` lock when checking for VMs with duplicate MAC-addresses.

fkorotkov: Oh. This comment was left from before when I tried generating in a different place. `needToGenerateNewMac` actually filters out by name as well so we should be good.

Sources/tart/Commands/Push.swift
```diff
@@ -74,7 +74,9 @@ struct Push: AsyncParsableCommand {
         // Populate the local cache (if requested)
         if populateCache {
           let expectedPushedVMDir = try ociStorage.create(pushedRemoteName)
-          try localVMDir.clone(to: expectedPushedVMDir, generateMAC: false)
+          // there might be a case when two remote VMs have identical MAC addresses
```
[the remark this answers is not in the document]
fkorotkov: Oh. My bad. I though we populating local image from remote. Need more coffee. 😅 Reverted in [74705ae](https://github.com/cirruslabs/tart/pull/539/commits/74705ae33ebb9bbab4b36697fb6ad6cc04f3d16a)

Sources/tart/Commands/Run.swift
```diff
@@ -112,7 +112,18 @@ struct Run: AsyncParsableCommand {
 
   @MainActor
   func run() async throws {
-    let vmDir = try VMStorageLocal().open(name)
+    let localStorage = VMStorageLocal()
+    let vmDir = try localStorage.open(name)
+    let needToGenerateNewMac = try localStorage.list().contains {
```
edigaryev, line 117: We should probably guard this logic with:

1. VM's `state()` check, to only generate new MAC-addresses for suspended VMs and balance with the same check in `tart clone`.
2. A lock, see https://github.com/cirruslabs/tart/pull/539#discussion_r1254420932.

Sources/tart/Commands/Run.swift
```diff
@@ -112,7 +112,23 @@ struct Run: AsyncParsableCommand {
 
   @MainActor
   func run() async throws {
-    let vmDir = try VMStorageLocal().open(name)
+    let localStorage = VMStorageLocal()
+    let vmDir = try localStorage.open(name)
+
+    let storageLock = try FileLock(lockURL: localStorage.baseURL)
```
edigaryev, line 118: Is there a reason to lock `localStorage.baseURL` instead of `Config().tartHomeDir` here?

Take a look at `Import.swift` and `Clone.swift` — they both use `Config().tartHomeDir` when modifying a list of VMs/checking for duplicated MAC-addresses.

Sources/tart/Commands/Run.swift
fkorotkov: Fixed in [ed4b3e1](https://github.com/cirruslabs/tart/pull/539/commits/ed4b3e16d96196d325bc62bfb28211c30a83e460)

edigaryev (approved):

Issue #139, opened 2022-06-27
Passthrough USB devices


fkorotkov:
We need to investigate how to passthough an iPhone connected to the host. This will be useful for running on device integration tests.

edigaryev:
Doesn't seem to be possible at this point, see https://github.com/docker/for-mac/issues/5263.

Doing something similar to [VirtualHere](https://www.virtualhere.com/) is an option, though, but most involves (1) delving into, most likely, [IOUSBHost](https://developer.apple.com/documentation/iousbhost) and [IOKit](https://developer.apple.com/documentation/iokit) frameworks and (2) installing the agent on each running VM that will forward the USB data packets to and from the host.

Thinking [of `adb`](https://stackoverflow.com/questions/23430181/adb-equivalent-for-ios-device), there's [an `idb` alternative](https://fbidb.io/) for iOS which [can serve emulators and devices](https://fbidb.io/docs/commands#starting-a-grpc-server) over the gRPC, but I'm not sure if Tart can directly help with doing that besides starting a VM and helping to get it's IP address.

fkorotkov:
Let's keep the issue open for visibility. `not possible atm` label should do fine.

geerlingguy:
This can also be useful for things like passing through block devices directly, or in my case passing an optical drive straight through to a VM for isolation/testing purposes.

sheerun:
What about employing https://github.com/freedesktop/spice-usbredir ?

dinvlad:
It’d be preferable to use something like usb-ip (but for Mac, as mentioned above) because (unlike USB redirection/pass-through solutions) it would allow sharing one device with multiple guests, from what I understand.

fkorotkov:
@geerlingguy it is now possible to pass block devices to that VMs via `--disk` flag. You need to run run tart under `sudo` though:

```
sudo TART_HOME="$HOME/.tart" tart run sonoma-xcode --disk=/dev/disk4s1
```

In the example above `/dev/disk4s1` is a USB drive that I have on my machine.

ink-splatters:
> @geerlingguy it is now possible to pass block devices to that VMs via `--disk` flag. You need to run run tart under `sudo` though:
> 
> ```
> sudo TART_HOME="$HOME/.tart" tart run sonoma-xcode --disk=/dev/disk4s1
> ```
> 
> In the example above `/dev/disk4s1` is a USB drive that I have on my machine.

[...]

PR #540, opened 2023-07-06
Resume suspended VMs without explicit `--suspendable` flag

fkorotkov:


edigaryev (approved):

Jon889:
@fkorotkov is there a way to have a VM run ephemerally in that it will restore from the saved state but not save the state at the end. Then the next time the VM runs it will restore from the same/original saved state

fkorotkov:
@Jon889 you can clone a VM into a temporary one which is very fast and will clone the state. Then you can run this temporary VM and delete it once not needed anymore.

Jon889:
Ah I thought that would invalidate the snapshot, as then multiple machines could be run. Also is it not possible to use snapshots with mounted directories? I get `Failed to suspend the VM: Invalid virtual machine configuration. Unsupported directory sharing device in configuration.` in which case it seems like tart shouldn't allow --suspendable and --dir to be used together

fkorotkov:
Sees you can resume a suspended VM with a mounted directory though. So you can run a suspendable VM without a mount, configure it and suspend. Then resume it again but this time with a mounted directory. The only part that is failing is the actual suspending of a VM with a mounted directory.

Jon889:
@fkorotkov Do you think there is any way around not being able to use suspendible VMs with mounted directories? macOS should have the ability to do it as USB drives can be plugged in and out without any notice.

edigaryev:
> @fkorotkov Do you think there is any way around not being able to use suspendible VMs with mounted directories? macOS should have the ability to do it as USB drives can be plugged in and out without any notice.

Unfortunately `Virtualization.Framework` does not support mounting USB drives yet, see https://github.com/cirruslabs/tart/issues/139 and https://github.com/utmapp/UTM/issues/3778 for more details.

[Mounting a block device](https://github.com/cirruslabs/tart/pull/611) could be an option, but a quick test shows that it also works incorrectly because the guest seems to be overwriting the host's device state/file table on resume from snapshot.

PR #541, opened 2023-07-06
Support stopping of suspended VMs

fkorotkov:


edigaryev:
I think that this logic should kick in in the `tart run` when the VM is stopped without suspending.

Otherwise the following scenario is possible:

1. `tart run --suspendable`
2. Suspend the VM
3. Run the VM, it will restore it's state and run just fine
4. Modify VM's disk/NVRAM state
5. Stop the VM
6. Run the VM again, this time it will fail after restoring it's state because it was created against earlier versions of disk/NVRAM

edigaryev (approved):

Issue #444, opened 2023-03-13
Feature request: macOS image downloading resume


roman-mishchenko:
Hi there!
During the cloning of the macOS image, I faced Internet connection problems. Thus, I had to restart the `tart clone ...` process again. So, I think it would be nice to have the resume option 🙂

edigaryev:
>So making tart pull resumable will involve 3 things:

What would be the criteria for garbage-collecting these "pull in-progress" directories?

I've had a simpler approach in mind, but it's simplicity again depends on the garbage-collection criteria:

1. Implement a cache that has roughly the following API (it will store blobs in `$TART_HOME/cache/blobs`, for example):

    ```swift
    protocol BlobCache {
      func get(digest: String) throws -> Data?
      func set(contents: Data) throws
    }
    ```

2. Check the cache before calling `Registry.pullBlob()`, if not `nil` — read the blob from the cache instead of retrieving it over the network.

4. Populate the cache in the `Registry.pullBlob()`'s callback (`handler` argument).

fkorotkov:
Any other command but `tart pull` will garbage collect it. If you tried to pull but failed we assume you will try to pull right after. If something like `tart clone` will be invoked, then it will garbage collect the partially pulled image.

I was also thinking about something like your `BlobCache` but it will require loading all layers first and only then uncompressing them which might be not ideal I think. Naive approach will double disk usage while pulling and to optimize it I think code will be a bit hard to understand.

Tommygithubaccount123:
I have [messaged](https://github.com/Tommygithubaccount123/tart/commit/c5a49e985d2030b9dfd9841bc0dbc8a0107cdbd2#commitcomment-121535579) @fkorotkov and it seems like [this method](https://github.com/cirruslabs/tart/issues/444#issuecomment-1622210092) is too complex. 

I will try to implement @edigaryev's [method](https://github.com/cirruslabs/tart/issues/444#issuecomment-1622249457)

PR #547, opened 2023-07-10
Make tmp dir deterministic

Tommygithubaccount123:
- Added temporary_deterministic func which takes in a fileName to name the tmpVM file

- tmpVMDir now calls the temporary_deterministic function

This is for: [step 1](https://github.com/cirruslabs/tart/issues/444#issuecomment-1622210092)

### DIFF
--- a/Sources/tart/VMDirectory.swift
+++ b/Sources/tart/VMDirectory.swift
@@ -58,20 +58,20 @@ struct VMDirectory: Prunable {
 
     return VMDirectory(baseURL: tmpDir)
   }
-  
+
   static func temporary_deterministic(fileName: String) throws -> VMDirectory {
     let fileName_noSlash = fileName.replacingOccurrences(of: "/", with: "-")
     let tmpDir = try Config().tartTmpDir.appendingPathComponent(fileName_noSlash)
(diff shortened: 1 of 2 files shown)


Sources/tart/VMDirectory.swift
```diff
@@ -59,6 +59,19 @@ struct VMDirectory: Prunable {
     return VMDirectory(baseURL: tmpDir)
   }
 
+  static func temporary_deterministic(fileName: String) throws -> VMDirectory {
+    let fileName_noSlash = fileName.replacingOccurrences(of: "/", with: "-")
```
fkorotkov, line 63: FYI Swift is using camel case.

fkorotkov:
There is no need to extract preliminary steps in separate PRs because they are not adding value on their own. These changes make sense only in the context of the resumable pull.

Tommygithubaccount123:
I was thinking to make the tmpVM's have a name that I could keep track of, like "ventura". Then if the "ventura" download was paused, the next PR could continue downloading in the "ventura" path.

I am assuming that all PRs will be resumable. What implementation were you thinking of?

PR #549, opened 2023-07-11
Fix `prune --older-than` deleting all cache

sjchmiela:
I noticed when I ran `prune --older-than=7` `tart` deleted some images I didn't expect it to (because I used them lately). I looked into `Prune` and it seems the sign is wrong on the `olderThan` argument handling.

## Before

Zrzut ekranu 2023-07-11 o 10 26 37

## After

Zrzut ekranu 2023-07-11 o 10 33 58

### DIFF
--- a/Sources/tart/Commands/Prune.swift
+++ b/Sources/tart/Commands/Prune.swift
@@ -34,7 +34,9 @@ struct Prune: AsyncParsableCommand {
     // Clean up cache entries based on last accessed date
     if let olderThan = olderThan {
       let olderThanInterval = Int(exactly: olderThan)!.days.timeInterval
-      let olderThanDate = Date().addingTimeInterval(olderThanInterval)
+      // olderThanInterval is positive (eg. 7 days) - we need to _subtract_
+      // the interval from current time to get "older than" date - hence * -1.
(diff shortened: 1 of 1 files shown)


Sources/tart/Commands/Prune.swift
```diff
@@ -34,7 +34,9 @@ struct Prune: AsyncParsableCommand {
     // Clean up cache entries based on last accessed date
     if let olderThan = olderThan {
       let olderThanInterval = Int(exactly: olderThan)!.days.timeInterval
-      let olderThanDate = Date().addingTimeInterval(olderThanInterval)
+      // olderThanInterval is positive (eg. 7 days) - we need to _subtract_
+      // the interval from current time to get "older than" date - hence * -1.
+      let olderThanDate = Date().addingTimeInterval(olderThanInterval * -1)
```
fkorotkov, line 39: Seems we can even more simplify the expression:

```suggestion
      let olderThanDate = Date() - olderThanInterval
```

edigaryev, line 39: Good catch, I think we can get away with just:

```suggestion
      let olderThanDate = Date() - olderThanInterval
```

fkorotkov (approved):

edigaryev (approved):

Issue #542, opened 2023-07-07
Allow deleting local images older than N days


sjchmiela:
The way I understand it, `tart prune` right now only allows pruning [remote images and IPSW cache](https://github.com/cirruslabs/tart/blob/e8dbb86fc097eb2e2712a1391983703285fc24b6/Sources/tart/Commands/Prune.swift#L124), _not_ local images.

Would it make sense to allow pruning local images too? Say, I'm building images on every commit in CI and after I build a fresh one I want to delete all older than 3 days (that aren't running, obviously).

It seems `VMStorageLocal` would only need to implement `PrunableStorage` (filtering out running ones) and `VMStorageLocal` would need to `updateAccessDate` in [`open`](https://github.com/cirruslabs/tart/blob/e8dbb86fc097eb2e2712a1391983703285fc24b6/Sources/tart/VMStorageLocal.swift#L14) (as does its OCI equivalent)?

What do you think?

fkorotkov:
I guess we can add something like `--source` argument similar to what `tart list` has. By default it will be equal to `oci`. @edigaryev do you have any thoughts on that?

edigaryev:
@fkorotkov I think that `tart prune` is currently somewhat cache-specific, perhaps a better and a more logical command for this would be `tart delete`?

fkorotkov:
Agreed, it might also complicate things when we'll add caches for resumable `tart pull`s. 🤔 

@sjchmiela can you maybe include something like date in the name of the temporary VMs and clean them via a script? Here is a script that will keep only 3 the most recent VMs (change `+4` in `tail` to control the amount) and will delete everything else.

```bash
tart list --source=local --format=json | jq --raw-output '.[] | .Name' | sort -r | tail -n +4  | xargs tart delete
```

sjchmiela:
Hmm, yea, maybe we can use that. My suggestion:
- uses access date, not create date (as using timestamp in name would), probably this doesn't matter in this CI/CD case,
- allows removing images by age more easily (arguably, still possible manually in the removing script).

Thanks!

fkorotkov:
Do I understand it correctly that you only build a VM once and push it right away to be consumed in a separate step right after? In that case it seems you just want to be able to see if a particular image is available remotely and in case of a "miss" build it and push it. And then you can delete it right away.

[...]

PR #550, opened 2023-07-11
Expose `AccessDate` in `tart list --format json`

sjchmiela:
Implements suggestion from https://github.com/cirruslabs/tart/issues/542#issuecomment-1629232417

### DIFF
--- a/Sources/tart/Commands/List.swift
+++ b/Sources/tart/Commands/List.swift
@@ -8,6 +8,7 @@ fileprivate struct VMInfo: Encodable {
   let Size: Int
   let Running: Bool
   let State: String
+  let AccessDate: Date?
 }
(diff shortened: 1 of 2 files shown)


fkorotkov:
As an alternative let's just support `tart delete --older-than`  in a separate PR. So you don't even need to parse the JSON at all for your use case. Note that `VMStorageLocal#open` method needs to update the access time similar to `VMStorageOCI#open`.

sjchmiela:
So actually after reworking our flow a bit (thanks to your suggestions and the remote image manifest upload time annotation) we ended up not needing this PR at all. And your suggestion here was implemented in https://github.com/cirruslabs/tart/pull/557, so… closing!

PR #551, opened 2023-07-11
Add OCI annotation with upload time

sjchmiela:
Adds an `org.cirruslabs.tart.upload-time` annotation to manifests pushed to OCI registries. Its value is current time (at the time of the `OCIManifest` object creation, ISO8601-formatted.

### DIFF
--- a/Sources/tart/OCI/Manifest.swift
+++ b/Sources/tart/OCI/Manifest.swift
@@ -5,6 +5,7 @@ let ociConfigMediaType = "application/vnd.oci.image.config.v1+json"
 
 // Annotations
 let uncompressedDiskSizeAnnotation = "org.cirruslabs.tart.uncompressed-disk-size"
+let uploadTimeAnnotation = "org.cirruslabs.tart.upload-time"
 
 struct OCIManifest: Codable, Equatable {
   var schemaVersion: Int = 2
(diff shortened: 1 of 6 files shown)


fkorotkov:
I think there is a conceptual difference for `created` field in case of containers and Tart VMs. With VMs you can run them and modify the state on disk. Container images are immutable by their nature. So I'm not sure if adding a concept of creation time is right for `VMConfig`. It seem the time there should be the last time this VM was stopped or suspended...

What do you and @edigaryev think of only adding `created` to the remote OCI manifest and populate it with the timestamp of when `tart push` was invoked?

edigaryev:
>What do you and @edigaryev think of only adding created to the remote OCI manifest and populate it with the timestamp of when tart push was invoked?

I don't mind putting any metadata in the OCI manifest as long it's done similarly to the [`org.cirruslabs.tart.uncompressed-disk-size` annotation](https://github.com/cirruslabs/tart/blob/826e5086463658c44dfe47c0fce2533a92cb730c/Sources/tart/OCI/Manifest.swift#L7).

sjchmiela:
> I think there is a conceptual difference for created field in case of containers and Tart VMs.

Ah, you're right. (Sorry, I keep looking at it from our perspective.)

> What do you and @edigaryev think of only adding `created` to the remote OCI manifest and populate it with the timestamp of when `tart push` was invoked?

That would work for me! I'll make this change in this pull request.

> I don't mind putting any metadata in the OCI manifest as long it's done similarly to the [`org.cirruslabs.tart.uncompressed-disk-size` annotation](https://github.com/cirruslabs/tart/blob/826e5086463658c44dfe47c0fce2533a92cb730c/Sources/tart/OCI/Manifest.swift#L7).

Hm, I could see `org.cirruslabs.tart.upload-time` too!

fkorotkov:
Then I think we all agreed on `org.cirruslabs.tart.upload-time` annotation. 🤝

fkorotkov (approved):

Sources/tart/OCI/Manifest.swift
```diff
@@ -5,6 +5,7 @@ let ociConfigMediaType = "application/vnd.oci.image.config.v1+json"
 
 // Annotations
 let uncompressedDiskSizeAnnotation = "org.cirruslabs.tart.uncompressed-disk-size"
+let uploadTimeAnnotation = "org.cirruslabs.tart.upload-time"
```
edigaryev, line 8: Looks more consistent with the rest of the code. What do you think?

Will also need to rename the `uploadTimeAnnotation` accordingly.

```suggestion
let uploadTimeAnnotation = "org.cirruslabs.tart.upload-date"
```

sjchmiela: Not sure to who is the question directed, but my 2 cents on my reasoning:
- Docker (and OCI spec) have `created` attribute. I considered `org.cirruslabs.tart.uploaded`, but it didn't seem right (I felt as if it could also be a boolean)
- I considered `….uploaded-at` (which would be the clearest to me), but it felt like breaking out of all conventions
- between `upload-time` and `upload-date` I picked `upload-time`, because I thought it hints at the fact it's not just a date (and the `Date` suffix in Swift variables are, to me, borrowed from `Date` class name, not name of the datetime concept per se).

I don't have strong opinion either way!

edigaryev (approved):

PR #553, opened 2023-07-11
Reclaim disk space only in the very end of `tart clone`

fkorotkov:
Please check comments in code for reasoning.


edigaryev:
What do you think about catching the specific out-of-disk-space exception on `clone()` instead?

This way we can be 100% sure that we need more disk space and we'll do a reclaim and then re-try.

fkorotkov:
I don't believe there will be any "specific out-of-disk-space exception on `clone()`" at all due to copy-on-write. 🤔

fkorotkov:
Alternatively we can monitor disk space on `tart run` and proactively claim space back when it's getting low. Not sure if there are macOS APIs available to get notified on low disk space.

PR #556, opened 2023-07-12
Don't delete an initiator of pruning

fkorotkov:
Sometimes people have an image that is greater than half of the disk itself. In that case such image will be pulled and prunned right away.

This change makes sure that an image that is being cloned from is not pruned right away.


fkorotkov:
@edigaryev I think you suggested something like that at some point.

edigaryev (dismissed):

fkorotkov:
Just realized we need to resolve symlinks [fbee31d](https://github.com/cirruslabs/tart/pull/556/commits/fbee31d1172f5dfca739577a304e3acdb275c344)

edigaryev (approved):

PR #560, opened 2023-07-15
Retrieve VM's IP for use in VNC after the VM is started

edigaryev:
Currently `tart run --vnc <name>` doesn't work because the `ScreenSharingVNC` tries to retrieve the VM's IP first: https://github.com/cirruslabs/tart/blob/4ce06279ff2376837c65933870ddb4a356d5f50e/Sources/tart/VNC/ScreenSharingVNC.swift#L14

...but at this point the VM is not started yet, so `tart run` locks up waiting for a miracle (which does happen when a VM with such MAC-address was already started on the host and the DHCPD lease was created for it).

### DIFF
--- a/Sources/tart/Commands/Run.swift
+++ b/Sources/tart/Commands/Run.swift
@@ -198,6 +198,17 @@ struct Run: AsyncParsableCommand {
       }
     }()
 
+    if let vncImpl = vncImpl {
+      vncImpl.setUserNotifier { vncURL in
+        if noGraphics || ProcessInfo.processInfo.environment["CI"] != nil {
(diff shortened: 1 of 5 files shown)


fkorotkov:
Should we just split `VM#run` into two methods? For example, move `if shouldResume` statement to `run` command all together, do `network.run(sema)` in both `start` and `resume` and the rest the current `run` method move into `VM#wait` method.

edigaryev:
> Should we just split `VM#run` into two methods?

Good idea, see https://github.com/cirruslabs/tart/pull/562.

PR #561, opened 2023-07-17
resuming of downloads

Tommygithubaccount123:
How it works:
- The first pull of a VM creates a progress file and a blobs-directory in the temporary-VM directory /.tart/tmp/VM-NAME
- A counter is used to track which diskLayers have been downloaded
- Blobs are now stored in the temporary blobs directory
- A progress file uses the counter to keep track of the progress
- Once all blobs have been pulled, the data is uncompressed/written

In a case of a stop:
- During a stop during downloading blobs, the process resumes at the blob that has not been downloaded
- During a stop during writing, the process restarts the entire writing process (does not need to pull blobs over network again)

[issue#444](https://github.com/cirruslabs/tart/issues/444)


Sources/tart/VMDirectory+OCI.swift
```diff
@@ -70,33 +74,61 @@ extension VMDirectory {
...
+    //diskCount keeps track of diskLayer
+    var diskCount = 0
+
+    //loop pulls one blob at a time
     for diskLayer in diskLayers {
-      try await registry.pullBlob(diskLayer.digest) { data in
-        try filter.write(data)
-        progress.completedUnitCount += Int64(data.count)
+      diskCount += 1
+      if !progFile.isDiskLayerDownloaded(diskLayer: diskCount){
```
edigaryev, line 87: 1. Do we need a progress file if we can simply check for the existence of a blob file on the disk?

2. Can we build the `progFile` progress on the fly, without using additional state? Simply increment  the `completedUnitCount` on each successful layer download/cache hit, that's it.

Also, since we now we have two stages: downloading and extracting, it probably makes sense to rename the `progFile` into `progressDownload` and `progress` into `progressExtract` and not forget to use the `ProgressObserver` on them.

Tommygithubaccount123:
- Now checks if the blob exists locally instead of using progress file

- added progressExtract/progressDownload

fkorotkov:
After some discussions it seems it's time to change how we store Tart images remotely. Even with this change we'll need to overcommit on disk usage: download everything first and unarchive next.

In #569 I'm proposing a new way of storing the images which will allow to accomplish everything: parallel resumable pulls with no extra disk consumption.

@Tommygithubaccount123 unfortunately from this PR we can only leave the deterministic temporary directory and GC avoidance for clone/pull. And for #569 me and @edigaryev will take over.

PR #570, opened 2023-07-28
GC avoidance and tmpDeterminstic

Tommygithubaccount123:
- avoid GC when command is a pull or clone
- make tmp VM names deterministic

From [this](https://github.com/cirruslabs/tart/pull/561#issuecomment-1653933742)

### DIFF
--- a/Sources/tart/VMDirectory.swift
+++ b/Sources/tart/VMDirectory.swift
@@ -59,6 +59,19 @@ struct VMDirectory: Prunable {
     return VMDirectory(baseURL: tmpDir)
   }
 
+  static func temporaryDeterministic(fileName: String) throws -> VMDirectory {
+    let fileNameNoSlash = fileName.replacingOccurrences(of: "/", with: "-")
+    let tmpDir = try Config().tartTmpDir.appendingPathComponent(fileNameNoSlash)
+    do {
+      try FileManager.default.createDirectory(at: tmpDir, withIntermediateDirectories: false)
+    }
+    catch{
+      defaultLogger.appendNewLine("\(fileNameNoSlash) exists")
+    }
+
+    return VMDirectory(baseURL: tmpDir)
+  }
+
   var initialized: Bool {
     FileManager.default.fileExists(atPath: configURL.path) &&
       FileManager.default.fileExists(atPath: diskURL.path) &&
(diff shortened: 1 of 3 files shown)


Sources/tart/VMDirectory.swift
```diff
@@ -59,6 +59,19 @@ struct VMDirectory: Prunable {
     return VMDirectory(baseURL: tmpDir)
   }
 
+  static func temporaryDeterministic(fileName: String) throws -> VMDirectory {
+    let fileNameNoSlash = fileName.replacingOccurrences(of: "/", with: "-")
+    let tmpDir = try Config().tartTmpDir.appendingPathComponent(fileNameNoSlash)
+    do {
+      try FileManager.default.createDirectory(at: tmpDir, withIntermediateDirectories: false)
```
fkorotkov, line 66: I think you can pass `true` to `withIntermediateDirectories` and don't catch the exception.

Sources/tart/VMStorageOCI.swift
```diff
@@ -170,7 +170,7 @@ class VMStorageOCI: PrunableStorage {
 
     if !exists(digestName) {
       let transaction = SentrySDK.startTransaction(name: name.description, operation: "pull", bindToScope: true)
-      let tmpVMDir = try VMDirectory.temporary()
+      let tmpVMDir = try VMDirectory.temporaryDeterministic(fileName: name.namespace)
```
fkorotkov, line 173: Seems either `digestName` or even full `name.description`. `name.description` contains slashes and might not be file safe so I think in this case you can rename parameter of `temporaryDeterministic` from `fileName` to something like `key` and hash the `key` using [CryptoKit](https://developer.apple.com/documentation/cryptokit/). Some insecure hashing like md5 will work just fine.

Sources/tart/VMStorageOCI.swift
```diff
@@ -170,7 +170,7 @@ class VMStorageOCI: PrunableStorage {
 
     if !exists(digestName) {
       let transaction = SentrySDK.startTransaction(name: name.description, operation: "pull", bindToScope: true)
-      let tmpVMDir = try VMDirectory.temporary()
+      let tmpVMDir = try VMDirectory.temporaryDeterministic(key: name.namespace)
```
fkorotkov, line 173: I think name doesn't contain `sha` and can cause a collision. Since we do hashing you can use the full description:

```suggestion
      let tmpVMDir = try VMDirectory.temporaryDeterministic(key: name.description)
```

fkorotkov (approved):

Issue #569, opened 2023-07-27
Improved format for fast and efficient pulls from remote OCI-registry


fkorotkov:
Right now in order to push an image Tart creates a single stream of compressed bytes and sends them in chunks one after another. Under the hood compressor has a state and in general data is compressed in "chunks" aka frames. With the current implementation it's possible that Tart will upload one layer ending with a compressed frame that only completes in the next layer.

This limitation requires to write the disk files sequentially in `tart pull` which makes parallel downloads generally impossible. At best we can only download layers in parallel but we'll need to unarchive them sequentially.

I propose to change this logic to the following. Each layer pushed to an OCI registry should contain:

1. Only completed compressed frames e.g. each layer should be independent and be unarchivable separatly.
2. Contain meta information about final uncompressed data:
    - Uncompressed size
    - Offset in the final file (can also be calculated on the fly from other layers)
    - checksum of the uncompressed 

This change will make life easier. We can implement resumable uploads pimplier than #527: upon pull for each layer we can see if checksum in `disk.img` at a given offset is equal to what we expect. This way we don't need to store any state of the donwload.

Having the layers being able to unarchive independently will allow to parallel downloads and unarchiving process.

Unfortunately, this new format will be incompatible with the current one. Images pushed with the v2 config format won't be consumable by the current Tart version. We'll need to mark this version Tart 2.0.0.

PR #589, opened 2023-08-18
Improved format for fast and efficient pulls from remote OCI-registry

edigaryev:
This turned out to be more complicated that I've expected, so any hints to simplify this are welcome.

Things that still need to be fixed and checked:

* ~~apply DRY to the logic of temporary files creation in `LayerizerTests` and implement their cleanup~~ → https://github.com/cirruslabs/tart/pull/589/commits/b4b79d390600ce9a994fd3675c6775695943fb2d
* ~~check if the `DiskV2.pull()` is really parallelizing pulls and not blocking due to e.g. `Digest.hash()`~~ → https://github.com/cirruslabs/tart/pull/589#issuecomment-1686848822 + https://github.com/cirruslabs/tart/pull/589#issuecomment-1686855945
* ~~check that the resumable pulls indeed work and maybe add an integration test for that~~ → https://github.com/cirruslabs/tart/pull/589#issuecomment-1687697649
* ~~check if we need to lock the `Progress` when updating it from `push()`'es and `pull()`'s~~ — we only modify [`Progress`](https://developer.apple.com/documentation/foundation/progress) from within a structured concurrency environment, and it is [`Sendable`](https://developer.apple.com/documentation/swift/sendable), so no problem here

Resolves https://github.com/cirruslabs/tart/issues/569, resolves https://github.com/cirruslabs/tart/issues/444.

### DIFF
--- a/Sources/tart/Commands/Push.swift
+++ b/Sources/tart/Commands/Push.swift
@@ -23,8 +23,8 @@ struct Push: AsyncParsableCommand {
                              """))
   var chunkSize: Int = 0
 
-  @Flag(help: .hidden)
-  var oldDiskFormat: Bool = false
+  @Option(help: .hidden)
+  var diskFormat: String = "v2"
 
   @Flag(help: ArgumentHelp("cache pushed images locally",
                            discussion: "Increases disk usage, but saves time if you're going to pull the pushed images later."))
@@ -73,7 +73,7 @@ struct Push: AsyncParsableCommand {
           registry: registry,
           references: references,
           chunkSizeMb: chunkSize,
-          oldDiskFormat: oldDiskFormat
+          diskFormat: diskFormat
         )
         // Populate the local cache (if requested)
         if populateCache {
--- a/Sources/tart/Commands/Push.swift
+++ b/Sources/tart/Commands/Push.swift
@@ -23,6 +23,10 @@ struct Push: AsyncParsableCommand {
                              """))
   var chunkSize: Int = 0
 
+  @Flag(help: ArgumentHelp("use the new V2 disk compression format when pushing this VM",
+                           discussion: "Pushing a VM with this flag enabled allows for additional concurrency when pulling this VM, see https://github.com/cirruslabs/tart/issues/569 for more details."))
+  var v2DiskFormat: Bool = false
+
   @Flag(help: ArgumentHelp("cache pushed images locally",
                            discussion: "Increases disk usage, but saves time if you're going to pull the pushed images later."))
   var populateCache: Bool = false
@@ -69,7 +73,8 @@ struct Push: AsyncParsableCommand {
         pushedRemoteName = try await localVMDir.pushToRegistry(
           registry: registry,
           references: references,
-          chunkSizeMb: chunkSize
+          chunkSizeMb: chunkSize,
+          v2DiskFormat: v2DiskFormat
         )
         // Populate the local cache (if requested)
         if populateCache {
--- a/Sources/tart/Commands/Push.swift
+++ b/Sources/tart/Commands/Push.swift
@@ -23,9 +23,8 @@ struct Push: AsyncParsableCommand {
                              """))
   var chunkSize: Int = 0
 
-  @Flag(help: ArgumentHelp("use the new V2 disk compression format when pushing this VM",
-                           discussion: "Pushing a VM with this flag enabled allows for additional concurrency when pulling this VM, see https://github.com/cirruslabs/tart/issues/569 for more details."))
-  var v2DiskFormat: Bool = false
+  @Flag(help: .hidden)
+  var oldDiskFormat: Bool = false
 
   @Flag(help: ArgumentHelp("cache pushed images locally",
                            discussion: "Increases disk usage, but saves time if you're going to pull the pushed images later."))
@@ -74,7 +73,7 @@ struct Push: AsyncParsableCommand {
           registry: registry,
           references: references,
           chunkSizeMb: chunkSize,
-          v2DiskFormat: v2DiskFormat
+          oldDiskFormat: oldDiskFormat
         )
         // Populate the local cache (if requested)
         if populateCache {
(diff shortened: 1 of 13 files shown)


edigaryev:
>check if the DiskV2.pull() is really parallelizing pulls and not blocking due to e.g. Digest.hash()

I've had a hard time finding a significant difference between the `DiskV1.pull()` and `DiskV2.pull()` when profiling the `LayerizerTests` (`testDiskV1` and `testDiskV2` respectively) using "Swift Concurrency" profiling template in [Instruments](https://en.wikipedia.org/wiki/Instruments_(software)) and running Docker Registry both in a container and [using `go run`](https://gist.github.com/edigaryev/e91aaace0e51e89cd933413d7a1492c5) (to use the host's FS directly).

However, things started differentiating nicely after I've added some traffic shaping to the [Docker Registry](https://github.com/distribution/distribution) (in this case each pull request for a blob was limited to a 20 MB/s or 160 Mb/s):

```patch
diff --git a/registry/storage/blobserver.go b/registry/storage/blobserver.go
index 6392e355..70968fbe 100644
--- a/registry/storage/blobserver.go
+++ b/registry/storage/blobserver.go
@@ -3,11 +3,13 @@ package storage
 import (
        "context"
        "fmt"
+       "io"
        "net/http"
        "time"
 
        "github.com/distribution/distribution/v3"
        "github.com/distribution/distribution/v3/registry/storage/driver"
+       "github.com/fujiwara/shapeio"
        "github.com/opencontainers/go-digest"
 )
 
@@ -73,6 +75,29 @@ func (bs *blobServer) ServeBlob(ctx context.Context, w http.ResponseWriter, r *h
                w.Header().Set("Content-Length", fmt.Sprint(desc.Size))
        }
 
-       http.ServeContent(w, r, desc.Digest.String(), time.Time{}, br)
+       http.ServeContent(w, r, desc.Digest.String(), time.Time{}, NewShapedReadSeeker(br, 20*1024*1024))
        return nil
 }
+
+type ShapedReadSeeker struct {
+       unshaped io.ReadSeeker
+       shaped   *shapeio.Reader
+}
+
+func NewShapedReadSeeker(unshaped io.ReadSeeker, bytesPerSec float64) *ShapedReadSeeker {
+       shaped := shapeio.NewReader(unshaped)
+       shaped.SetRateLimit(bytesPerSec)
+
+       return &ShapedReadSeeker{
+               unshaped: unshaped,
+               shaped:   shaped,
+       }
+}
+
+func (srs *ShapedReadSeeker) Read(p []byte) (n int, err error) {
+       return srs.shaped.Read(p)
+}
+
+func (srs *ShapedReadSeeker) Seek(offset int64, whence int) (int64, error) {
+       return srs.unshaped.Seek(offset, whence)
+}
```

Here are the results for `testDiskV1` creating, pushing and pulling a 5 GB disk containing random data, without the final file comparison step:

Screenshot 2023-08-21 at 22 36 15

The total runtime was 4m50s, out of which the pulling time was 4m20s.

Results for `testDiskV2` creating, pushing and pulling a 5 GB disk containing random data, without the final file comparison step and using `concurrency: 4` (instead of the default `concurrency: 16` in this test):

Screenshot 2023-08-21 at 22 40 39

The total runtime was 1m46s, out of which the pulling time was 1m20s, which is 325% faster than the `testDiskV1` result.

edigaryev:
>check if the DiskV2.pull() is really parallelizing pulls and not blocking due to e.g. Digest.hash()

To check how the `Digest.hash()` behaves, I've duplicated the `DiskV2.pull()` call one more time in `testDiskV2`, here's a profile for pull task groups doing only `Digest.hash()` on a file range:

Screenshot 2023-08-21 at 22 47 49

They seem to be getting parallellized.

edigaryev:
>check that the resumable pulls indeed work

I've manually checked this by:

1. Starting a pull for a `ghcr.io/cirruslabs/macos-ventura-base:latest` image re-pushed to use the V2 disk format.
2. Disabling the Wi-Fi to get the `Error: The network connection was lost.`.
3. Copying the `disk.img` from `ghcr.io/cirruslabs/macos-ventura-base:latest` into `~/.tart/tmp/<MD5>`

After that, the digest verification for the disk chunks finished in a matter of seconds and only NVRAM was downloaded.

One thing I've noticed that the `~/.tart/tmp/<MD5>` directory will be removed when exiting the `tart pull` using Ctrl+C. But I'm not sure if this is an issue because we're mainly targeting network errors with resumable pulls.

fkorotkov:
Rebased the PR to include #591 to run some tests on AWS network.

fkorotkov:
I did some testing on a `mac2.metal` instance in AWS with the latest `monterey-xcode` image (40GB). I pushed the image in both `v1` and v2` format to the ECR and tried to pull them respectively (plus experiment with `--concurrency`):

- `v1` pull showed time around `23:00`.
- `v2` pull showed `5:45` time
- **`v2 --concurrency=8` pull showed staggering `4:03` time**

Seems with the default concurrency of 4 we are getting exactly 4x faster pulls and trying to increase it further pushes it to the limit.

Sources/tart/Commands/Push.swift
```diff
@@ -23,6 +23,9 @@ struct Push: AsyncParsableCommand {
                              """))
   var chunkSize: Int = 0
 
+  @Flag(help: .hidden)
+  var oldDiskFormat: Bool = false
```
fkorotkov, line 27: I have some ideas for even further improvements so we might change the format again in a year or so. Let's make it a vertion thing like `v1`, `v2`, etc. This way it will be easier for us to modify Pakcer builds to still build ventura images using v1 and only sonoma using v2.

edigaryev: Implemented in https://github.com/cirruslabs/tart/pull/589/commits/e529472ad668ce56e71431d9ad86a991e1d10307.

fkorotkov (approved):

PR #606, opened 2023-09-21
Add Uphold as a Tart user

ruimarinho:
At Uphold, we use Tart to manage ephemeral GitHub Actions runners for building our iOS and Android applications and running end-to-end UI tests. Whenever a new webhook comes in, a listener interacts with [Orchard](https://github.com/cirruslabs/orchard) and launches a new Tart VM to run each job on a clean state. Along with Tart, we use the Actions Runner Controller (ARC) Kubernetes operator for other tasks. 

We're integrating PiKVM 4 into our Mac Mini fleet, which is positioned on 3D-printed supports sitting on a rack. Through MDM, we deploy an Ansible playbook automatically registering each machine as an Orchard Worker without any user interaction, and setting Tart VM resources (i.e. VM "slots") based on the physical capabilities of the hardware, primarily RAM.

We've been contributing with engagement to the Tart community and hope that in the future there will be an opportunity for us to also become a sponsor. We would like to extend our sincere thanks to the authors and contributors of Tart for their hard work and dedication.


SandroMachado (approved):

fkorotkov (approved):
Sounds like a very nice setup! 👏 Would love to learn about your MDM/Ansible setups more at some point.

PR #610, opened 2023-09-22
.cirrus.yml: add an execution_lock: for integration tests

edigaryev:
Resolves PERSISTENT-WORKERS-1A.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -4,13 +4,14 @@ env:
   XCODE_TAG: 15
 
 task:
-  execution_lock: "integration-tests"
   name: Test on Ventura
   alias: test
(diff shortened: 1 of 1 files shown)


fkorotkov (commented):
I wonder if we should resolve it by having resource labels on the worker? Theoretically we can run two integration tests at once. What do you think?

edigaryev:
> I wonder if we should resolve it by having resource labels on the worker? Theoretically we can run two integration tests at once. What do you think?

Indeed, added in https://github.com/cirruslabs/tart/pull/610/commits/53e65bf7e986dbaa389d380e956d2148b6600032.

fkorotkov (approved):

PR #621, opened 2023-10-03
Generalize a cache for downloading files

fkorotkov:
Related to #620


Sources/tart/DownloadsCache.swift
```diff
@@ -0,0 +1,69 @@
...
+    for try await chunk in channel {
+      let chunkAsData = Data(chunk)
+      fileHandle.write(chunkAsData)
+      progress.completedUnitCount += Int64(chunk.count)
+    }
+
+    try fileHandle.close()
+
+    if let etag = response.value(forHTTPHeaderField: "etag") {
+      let finalLocation = locationFor(fileName: "etag:\(etag).\(fileExtension)")
```
edigaryev, line 62: Should we trust the ETags returned by the HTTP servers on the internet? I don't think so.

Perhaps we should continue to use the old content-addressable storage approach here too.

fkorotkov: But the hash is different from the ETags. I can prefix the whole thing with a hash of the URL so if two URLs have the same etag it won't screw things up. What do you thing?

fkorotkov: Took a stab with [4433e65](https://github.com/cirruslabs/tart/pull/621/commits/4433e654a9eda18f28438fe6a6c8c65633d8708a). PTAL.

PR #643, opened 2023-11-01
Fail fast of can't locate registry credentials

fkorotkov:
If we are getting challenged by the registry to acquire an authentication token, then there should have credentials in place. Let's fail fast if there are no credentials.

### DIFF
--- a/Sources/tart/OCI/Registry.swift
+++ b/Sources/tart/OCI/Registry.swift
@@ -106,7 +106,9 @@ class Registry {
   var currentAuthToken: Authentication? = nil
 
   var host: String? {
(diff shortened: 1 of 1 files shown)


edigaryev:
> If we are getting challenged by the registry to acquire an authentication token, then there should have credentials in place.

Not necessarily. For example, when pulling `ghcr.io/cirruslabs/debian:latest` as an anonymous user, `https://ghcr.io/v2/` will return `HTTP 401`, but we'll still be able to authenticate against the presented realm even without the credentials provided (you can try this yourself by running `curl "https://ghcr.io/token?scope=repository:cirruslabs/debian:pull&service=ghcr.io"`).

PR #647, opened 2023-11-06
Introduce AuthenticationKeeper actor to serialize authn modification

edigaryev:
We also need to prevent multi-entrancy here:

https://github.com/cirruslabs/tart/blob/f4bc02d1751dd5ad447619515e7d73a9efc2d7e6/Sources/tart/OCI/Registry.swift#L313-L319

...but let's do that in a separate PR to avoid complicating things.


fkorotkov:
Also agreed on the multi-entrancy thing, don't think we even need to care of it since generating tokens should be handled by the registry. Registries should support parallel pull on their end.

PR #651, opened 2023-11-08
Properly enter main even loop in headless mode

### LINKED ISSUE #638: `--no-graphics` cause a white icon that keeps beating
Host: macOS 13.6
Guest: macOS 14.1
Tart: 2.1.0

When using `--no-graphics` headless flag, there is a white `Tart` icon will continue to bouncing in the dock until the VM is manually stopped.

https://github.com/cirruslabs/tart/assets/61939856/777df2ea-0f84-4245-a5d3-62e2177ab045

fkorotkov:
Fixes #638

### DIFF
--- a/Sources/tart/Commands/Run.swift
+++ b/Sources/tart/Commands/Run.swift
@@ -304,7 +304,8 @@ struct Run: AsyncParsableCommand {
 
     let useVNCWithoutGraphics = (vnc || vncExperimental) && !graphics
     if noGraphics || useVNCWithoutGraphics {
-      dispatchMain()
+      // enter the main even loop and just wait for the VM to exit
+      NSApplication.shared.run()
     } else {
       runUI(suspendable)
     }


edigaryev:
How you were able to validate this? Because for me even when running `Tart 2.2.1` installed via Homebrew there's no icon jumping in Docker on macOS Sonoma 14.1 (I'm running `ghcr.io/cirruslabs/ubuntu:latest`, though, will try macOS shortly).

Also, it would be nice to explain a bit on why this is the proper way of entering the main event loop and why it results in no icon jumping in Dock.

fkorotkov:
I reproduced with `tart run --no-graphics sonoma-xcode`, checked [`dispatchMain` documentation](https://developer.apple.com/documentation/dispatch/1452860-dispatchmain) which mentioned `NSApplicationMain`. Then I googled a bit and found another mention of the Application main loop [on StackOverflow](https://stackoverflow.com/questions/1006648/how-do-i-make-the-dock-icon-stop-bouncing-after-my-app-wrapper-script-starts/1008267#1008267).

Made this change, ran it locally and didn't see the bouncing box. Also tested that Control+C works as expected.

edigaryev (approved):

PR #654, opened 2023-11-10
Print put errors from Docker Helpers

fkorotkov:


Sources/tart/Credentials/DockerConfigCredentialsProvider.swift
```diff
@@ -41,13 +41,17 @@ class DockerConfigCredentialsProvider: CredentialsProvider {
 
     process.waitUntilExit()
 
+    let outputData = try outPipe.fileHandleForReading.readToEnd()
     if !(process.terminationReason == .exit && process.terminationStatus == 0) {
+      let outputText = outputData == nil ? "" : String(decoding: outputData!, as: UTF8.self)
+      print(outputText)
```
edigaryev, line 47: Perhaps we could simplify this down to:

```suggestion
      if let outputData = outputData {
        print(String(decoding: outputData, as: UTF8.self))
      }
```

Haven't checked this against compiler, though.

Sources/tart/Credentials/DockerConfigCredentialsProvider.swift
```diff
@@ -41,13 +41,17 @@ class DockerConfigCredentialsProvider: CredentialsProvider {
...
       throw CredentialsProviderError.Failed(message: "Docker helper failed!")
     }
+    if outputData == nil {
```
edigaryev, line 50: Could the `outputData` be non-nil, but still empty?

fkorotkov: Might be but it will mean that the helper binary breaks contract by not outputting anything. JSON decoding will break then.

edigaryev: But can we check for the `outputData?.count == 0` instead?

That will be much cleaner IMO 🤔

edigaryev (dismissed):

Sources/tart/Credentials/DockerConfigCredentialsProvider.swift
fkorotkov: As you wish! See [a637f10](https://github.com/cirruslabs/tart/commit/a637f1093f6698dbcba0844fae02b7c15cd7c23a) [now](https://github.com/cirruslabs/tart/pull/654#event-10928229153)

edigaryev (approved):

PR #675, opened 2023-11-30
Fix the filesystem corruption on Linux VMs

amalchuk:
Working on Linux virtual machine for a long-time causes to disk corruption:

```
bash: /usr/bin/ls: Input/output error
```

Same issues in another projects:

- https://github.com/utmapp/UTM/issues/4840
- https://github.com/lima-vm/lima/issues/1957

These changes:
- Uses NVMe as the default for Linux VMs on macOS 14+
- On macOS 12+ hosts, the caching mode for virtio drives in Linux VMs is changed to `.cached`

### DIFF
--- a/Sources/tart/VM.swift
+++ b/Sources/tart/VM.swift
@@ -326,9 +326,14 @@ class VM: NSObject, VZVirtualMachineDelegate, ObservableObject {
     }
 
     // Storage
-    var devices: [VZStorageDeviceConfiguration] = [
-      VZVirtioBlockDeviceConfiguration(attachment: try VZDiskImageStorageDeviceAttachment(url: diskURL, readOnly: false))
-    ]
+    let attachment: VZDiskImageStorageDeviceAttachment = try getStorageDevice(diskURL: diskURL, useFsWorkAround: vmConfig.os == .linux)
+    var device: VZStorageDeviceConfiguration
+    if #available(macOS 14, *), vmConfig.os == .linux {
+      device = VZNVMExpressControllerDeviceConfiguration(attachment: attachment)
+    } else {
+      device = VZVirtioBlockDeviceConfiguration(attachment: attachment)
+    }
+    var devices: [VZStorageDeviceConfiguration] = [device]
     devices.append(contentsOf: additionalStorageDevices)
     configuration.storageDevices = devices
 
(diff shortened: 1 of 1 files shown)


edigaryev:
Hello Andrew 👋  Thanks for spotting the issue and providing a fix!

I've looked at the links you've provided, and it seems that the lima project got this fixed [by just forcing the cached mode for the VirtIO disk](https://github.com/lima-vm/lima/pull/2026/files).

It also looks like the [NVMe device is actually slower](https://github.com/utmapp/UTM/pull/5919) than the VirtIO device, and glancing at the [172-page spec of NVM Express](https://nvmexpress.org/wp-content/uploads/NVM-Express-1_1b-1.pdf) it should be more complex too (compare this to 16-pages of the [VirtIO specification](https://docs.oasis-open.org/virtio/virtio/v1.3/virtio-v1.3.pdf) for block devices).

What do you think about skipping the NVMe part for now?

edigaryev (dismissed):

Sources/tart/VM.swift
```diff
@@ -326,9 +326,14 @@ class VM: NSObject, VZVirtualMachineDelegate, ObservableObject {
     }
 
     // Storage
-    var devices: [VZStorageDeviceConfiguration] = [
-      VZVirtioBlockDeviceConfiguration(attachment: try VZDiskImageStorageDeviceAttachment(url: diskURL, readOnly: false))
-    ]
+    let attachment: VZDiskImageStorageDeviceAttachment = try getStorageDevice(diskURL: diskURL, useFsWorkAround: vmConfig.os == .linux)
```
fkorotkov, line 329: Small nit-pick about passing `useFsWorkAround` since we [only support macOS 13+ already](https://github.com/cirruslabs/tart/blob/ac5f794e6d69652ee4841562c8c94e23df7ae6c2/Package.swift#L7). When looking at `getStorageDevice` alone it was not trivial for me to figure out when this workaround coming from. What do you think of inlining the method via ternary operator. IMO it improves readability. What do you think?

```suggestion
    let attachment: VZDiskImageStorageDeviceAttachment = vmConfig.os == .linux ?
      // Use cached caching mode for virtio drive to prevent fs corruption on linux when possible
      try VZDiskImageStorageDeviceAttachment(url: diskURL, readOnly: false, cachingMode: .cached, synchronizationMode: .full) :
      try VZDiskImageStorageDeviceAttachment(url: diskURL, readOnly: false)
```

Sources/tart/VM.swift
```diff
@@ -362,6 +367,15 @@ class VM: NSObject, VZVirtualMachineDelegate, ObservableObject {
...
 
+  private static func getStorageDevice(diskURL: URL, useFsWorkAround: Bool = false) throws -> VZDiskImageStorageDeviceAttachment {
+    if #available(macOS 12, *), useFsWorkAround {
+      // Use cached caching mode for virtio drive to prevent fs corruption on linux when possible
+      return try VZDiskImageStorageDeviceAttachment(url: diskURL, readOnly: false, cachingMode: .cached, synchronizationMode: .full)
+    } else {
+      return try VZDiskImageStorageDeviceAttachment(url: diskURL, readOnly: false)
+    }
+  }
+
```
fkorotkov, line 378: With my suggestion above we can remove this method:

```suggestion
```

Sources/tart/VM.swift
amalchuk: Added as you want. 🙂

fkorotkov (approved):

PR #694, opened 2024-01-03
tart set: bring back the --disk-size command-line argument

edigaryev:
See https://github.com/cirruslabs/linux-image-templates/pull/7.

### DIFF
--- a/Sources/tart/Commands/Set.swift
+++ b/Sources/tart/Commands/Set.swift
@@ -16,7 +16,18 @@ struct Set: AsyncParsableCommand {
   @Option(help: "VM display resolution in a format of <width>x<height>. For example, 1200x800")
   var display: VMDisplayConfig?
 
-  @Option(help: "Resize the VMs disk to the specified size in GB (note that the disk size can only be increased to avoid losing data")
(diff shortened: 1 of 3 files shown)


Sources/tart/Commands/Set.swift
```diff
@@ -16,7 +16,7 @@ struct Set: AsyncParsableCommand {
   @Option(help: "VM display resolution in a format of <width>x<height>. For example, 1200x800")
   var display: VMDisplayConfig?
 
-  @Option(help: .hidden)
+  @Option(help: "Resize the VMs disk to the specified size in GB (note that the disk size can only be increased to avoid losing data")
```
fkorotkov, line 19: Let's add a `discussion` section that explains that most of Linux distros should work out of the box but for macOS one needs to run some commands to resize it and make macOS aware of the change.

edigaryev: See https://github.com/cirruslabs/tart/pull/694/commits/42c24702d7517f89b8b5fbc78164d4912b52ee63.

fkorotkov (approved):

PR #716, opened 2024-01-22
Build x86 binary

fkorotkov:
So we can use it to build amd64 images for Vetu to use for [Cirrus Runners](https://cirrus-runners.app/).

`#if`s are ugly but I didn't find a better way. I wish Swift had something like `build` in Go.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,7 +1,7 @@
 use_compute_credits: true
 
 env: 
-  XCODE_TAG: 15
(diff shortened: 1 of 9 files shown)


edigaryev (approved):

edigaryev (approved):

edigaryev (changes requested):
I think we can now bring back the `tart create --from-ipsw` test now that the issue with networking is resolved.

edigaryev (approved):

PR #723, opened 2024-01-30
Improve rename error message

noahsmartin:
This seems like the intention of the error message, it should show the name of the VM that you are trying to set as the new name.

If you run:
`tart rename firstVM secondVM`

It currently prints:
`failed to rename VM firstVM, target VM firstVM already exists`

Now it will print:
`failed to rename VM firstVM, target VM secondVM already exists`


edigaryev (approved):

fkorotkov (approved):

fkorotkov:
Thank you, @noahsmartin! 🙌

Issue #305, opened 2022-11-06
Desktop Application


fkorotkov:
Right now Tart is specifically a CLI tool that is meant to be integrated into some other system or run from a terminal. And Tart is daemon-less, meaning that to stop a VM one need to kill `tart` process.

For more user friendly experience, let's consider creating `Tart.App` with some UI similar to what Parallels has:

<img width="394" alt="Screenshot 2022-11-05 at 8 18 47 PM" src="https://user-images.githubusercontent.com/989066/200148296-a0d5a136-32a6-4a50-a129-d698252b3299.png">

So a user can view a list of VMs, can change config, start and stop them, etc.

fkorotkov:
Just a quick shout out to @vitorgalvao for creating a [Alfredo workflow](https://alfred.app/workflows/vitor/virtual-machines/):

![image](https://user-images.githubusercontent.com/989066/208701011-7a215dd9-34b4-46a4-a5c8-f641dce3e571.png)

mafreud:
Great suggestion! If you use Flutter, I can contribute since I have some experiences of it.

PR #728, opened 2024-02-16
Add command line option to run without audio

torarnv:
The default behavior is to run with audio if we're showing the Tart UI window. It can be disabled with --no-audio.

When running without a UI Window, audio is disabled, but can be enabled via --audio.

### DIFF
--- a/Sources/tart/Commands/Run.swift
+++ b/Sources/tart/Commands/Run.swift
@@ -36,6 +36,9 @@ struct Run: AsyncParsableCommand {
   @Flag(help: "Force open a UI window, even when VNC is enabled.")
   var graphics: Bool = false
 
+  @Flag(help: "Disable audio pass-through to host.")
+  var noAudio: Bool = false
+
   #if arch(arm64)
     @Flag(help: "Boot into recovery mode")
   #endif
@@ -210,7 +213,8 @@ struct Run: AsyncParsableCommand {
       additionalStorageDevices: additionalDiskAttachments,
       directorySharingDevices: directoryShares() + rosettaDirectoryShare(),
       serialPorts: serialPorts,
-      suspendable: suspendable
+      suspendable: suspendable,
+      audio: !noAudio
     )
 
     let vncImpl: VNC? = try {
--- a/Sources/tart/VM.swift
+++ b/Sources/tart/VM.swift
@@ -45,7 +45,8 @@ class VM: NSObject, VZVirtualMachineDelegate, ObservableObject {
        additionalStorageDevices: [VZStorageDeviceConfiguration] = [],
        directorySharingDevices: [VZDirectorySharingDeviceConfiguration] = [],
        serialPorts: [VZSerialPortConfiguration] = [],
-       suspendable: Bool = false
+       suspendable: Bool = false,
+       audio: Bool = true
   ) throws {
     name = vmDir.name
     config = try VMConfig.init(fromURL: vmDir.configURL)
@@ -61,7 +62,8 @@ class VM: NSObject, VZVirtualMachineDelegate, ObservableObject {
                                                 network: network, additionalStorageDevices: additionalStorageDevices,
                                                 directorySharingDevices: directorySharingDevices,
                                                 serialPorts: serialPorts,
-                                                suspendable: suspendable
+                                                suspendable: suspendable,
+                                                audio: audio
     )
     virtualMachine = VZVirtualMachine(configuration: configuration)
 
@@ -274,7 +276,8 @@ class VM: NSObject, VZVirtualMachineDelegate, ObservableObject {
     additionalStorageDevices: [VZStorageDeviceConfiguration],
     directorySharingDevices: [VZDirectorySharingDeviceConfiguration],
     serialPorts: [VZSerialPortConfiguration],
-    suspendable: Bool = false
+    suspendable: Bool = false,
+    audio: Bool = true
   ) throws -> VZVirtualMachineConfiguration {
     let configuration = VZVirtualMachineConfiguration()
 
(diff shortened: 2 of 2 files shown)


Sources/tart/Commands/Run.swift
```diff
@@ -36,6 +36,16 @@ struct Run: AsyncParsableCommand {
...
 
+  @Flag(help: ArgumentHelp(
+    "Pass audio through to host.",
+    discussion: "This is the default if opening a UI window."))
+  var audio: Bool = false
+
+  @Flag(help: ArgumentHelp(
+    "Don't pass audio through to host.",
+    discussion: "Disables audio pass through, even when opening a UI window."))
+  var noAudio: Bool = false
```
fkorotkov, line 47: There is no need in `noAudio`. SwiftArguments will handle `--no-audio` if you have `--audio`. You can leave only `audio` variable and default it to `true` for backward compatibility.

Sources/tart/Commands/Run.swift
```diff
@@ -204,13 +218,18 @@ struct Run: AsyncParsableCommand {
       serialPorts.append(createSerialPortConfiguration(tty_read!, tty_write!))
     }
 
+    let useVNC = vnc || vncExperimental
+    let useBuiltInGraphics = (useVNC && graphics) || !noGraphics
+    let useAudio = (useBuiltInGraphics && !noAudio) || audio;
```
fkorotkov, line 223: `useVNC && graphics` seems a bit off. Can we just start passing `audio` down to `craftConfiguration` without changing these variables and logic.

fkorotkov (commented):
Thank you for the improvement! Great idea to allow control audio in the VM. IMO overthinking with separate `--no-audio` and `--audio` with custom logic for VNC/nonVNC is too complex and magical. What do you think of a single `audio` flag that will allow to directly control audio without customization for window/no window?

Sources/tart/Commands/Run.swift
torarnv: Ah, great! I modeled the two variables after `graphics` and `noGraphics`.

torarnv:
> Thank you for the improvement! Great idea to allow control audio in the VM. IMO overthinking with separate --no-audio and --audio with custom logic for VNC/nonVNC is too complex and magical. What do you think of a single audio flag that will allow to directly control audio without customization for window/no window?

I think the expectation when running with `--no-graphics` is that Tart is running as a service, where the host is not affected. Having random sounds come out of the host sound device because some audio based CI test in the guest is running seems strange with that expectation 😄 

If disabling audio in no-graphics mode is too magical, perhaps we should introduce a `--background` or `--daemon` or `--detach` flag, that disables both graphics and audio, and also backgrounds Tart instead of running in the foreground? It's something I've missed, similar to `docker start`, where I can kick off Tart in the background without fiddling with stuff like `nohup tart run something >/dev/null 2>&1 &`.

Sources/tart/Commands/Run.swift
fkorotkov: Oh. Maybe I confused SwiftArguments features with what we use for Go. Regardless, I think we can leave just one variable either `audio = true` or `noAudio = false`. And pass it down to `craftConfiguration`.

fkorotkov:
I did a little digging and `--graphics` was introduced in #248 to support having both VNC and UI for debugging Packer plugin in https://github.com/cirruslabs/packer-plugin-tart/pull/21 This way Packer plugin runs something like `tart run sonoma-base --graphics --vnc-experimental` to have both VNC for the plugin to send commands to and UI window for debugging by a person.

In case of audio still seems like a single `--no-audio` mimicking `--no-graphics` is sufficient enough.

I think in most of the cases when `--no-graphics` is used it's used on a host somewhere in a datacenter where either sounds on the host is already disabled or there are no people to bother. 😅 Introducing yet another flag like `--headless` is an option but might be too much for just passing `--no-graphics --no-audio` by an integrator.

torarnv:
> I did a little digging and `--graphics` was introduced in #248 to support having both VNC and UI for debugging Packer plugin in [cirruslabs/packer-plugin-tart#21](https://github.com/cirruslabs/packer-plugin-tart/pull/21) This way Packer plugin runs something like `tart run sonoma-base --graphics --vnc-experimental` to have both VNC for the plugin to send commands to and UI window for debugging by a person.

Interesting you should mention that :) One of my use-cases is that I want to run packer provisioning using Tart without any graphics _or audio_. Would you expect the `headless` option in the [Tart packer plugin](https://developer.hashicorp.com/packer/integrations/cirruslabs/tart/latest/components/builder/tart) to then map to both `--no-graphics` and `--no-audio` ?



> In case of audio still seems like a single `--no-audio` mimicking `--no-graphics` is sufficient enough.

Fair enough :)

> I think in most of the cases when `--no-graphics` is used it's used on a host somewhere in a datacenter where either sounds on the host is already disabled or there are no people to bother. 😅 Introducing yet another flag like `--headless` is an option but might be too much for just passing `--no-graphics --no-audio` by an integrator.

For the data center case, I'd agree. But what do you think of a `detach/daemon` option? Having to keep a terminal open (and remembering to not close it to not kill the VM) is a bit annoying :)

fkorotkov:
For the packer plugin we might just add an option to disable audio as well. Might be handy for when we use screen reader 😅 

`detach/daemon/headless` is a theme that was discussed several times on different occasions and we'd agreed in most cases that integrations should handle VM lifecycle and "background" activity. For example, [Cirrus CLI](https://tart.run/integrations/cirrus-cli/) can help you automate your testing. Or if your custom test runner can handle starting and stopping VMs.

We might reconsider developing Tart Desktop application #305 that will sit in your tray. But we need more than just one use case to allocate time for developing and maintaining it.

fkorotkov:
> Why not use the existing `headless` option?

Maybe it can be a part of `headless` but also as a separate option. It's something to think about when making a PR to the Packer plugin. My initial though is to have it configured separately since IMO headless is about no GUI rather than no nothing.  Anyways it's a discussion outside of the scope of this PR.

fkorotkov:
As to merge this PR let's just keep one `--no-audio` option to disable audio device. Does it sound fair?

torarnv:
Yepp, I'll re-push based on that.

fkorotkov (approved):

edigaryev (approved):

PR #731, opened 2024-02-19
Introduce `--dir-device-tag` flag

fkorotkov:
To override a tag for VirtioFS device.


edigaryev (approved):

torarnv:
I think this approach is strictly worse than the one in https://github.com/cirruslabs/tart/pull/730

PR #732, opened 2024-02-19
Do not magically set `--no-graphics` when `--vnc` is passed

fkorotkov:
From a discussion in #728 it appeared that having both `--graphics` and `--no-graphics` is a bit confusing.

`--graphics` was introduced in #248 to support having both VNC and UI for debugging Packer plugin in cirruslabs/packer-plugin-tart#21. This is because `--vnc` flag has a side effect of hiding UI which I think was wrong in retrospective. One can run `tart run --vnc --no-graphics`. In most of the cases this is automated via Alfred or something like that.

Now we have so many arguments that IMO it's worth to remove `--graphics` for overall consistency in arguments: everything is enabled by default and can be disabled via `--no-*` flags.

### DIFF
--- a/Sources/tart/Commands/Run.swift
+++ b/Sources/tart/Commands/Run.swift
@@ -33,7 +33,7 @@ struct Run: AsyncParsableCommand {
   ))
   var serialPath: String?
 
-  @Flag(help: "Force open a UI window, even when VNC is enabled.")
+  @Flag(help: ArgumentHelp("Force open a UI window, even when VNC is enabled.", visibility: .hidden))
   var graphics: Bool = false
 
   #if arch(arm64)
@@ -133,8 +133,8 @@ struct Run: AsyncParsableCommand {
       throw ValidationError("--net-bridged and --net-softnet are mutually exclusive")
     }
 
-    if graphics && noGraphics {
-      throw ValidationError("--graphics and --no-graphics are mutually exclusive")
+    if graphics {
+      print("--graphics is deprecated and will be removed in the future.\n")
     }
 
     if (noGraphics || vnc || vncExperimental) && captureSystemKeys {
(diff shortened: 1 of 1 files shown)


edigaryev (approved):

torarnv:
I noticed this new behavior today when trying the `--vnc` and `--vnc-experimental` options. 

The command line documentation still says "Use screen sharing instead of the built-in UI." and "Use Virtualization.Framework's VNC server instead of the build-in UI.". Should we update these?

Another complication is that we use `--no-graphics` to determine whether the VNC connection should be automatically opened or not:

```
         if noGraphics || ProcessInfo.processInfo.environment["CI"] != nil {
            print("VNC server is running at \(vncURL)")
          } else {
            print("Opening \(vncURL)...")
            NSWorkspace.shared.open(vncURL)
          }
```

Deprecating `--graphics` makes sense, but perhaps we should reconsider the behavior of  `--vnc` and `--vnc-experimental` in terms of opening the default UI? 

Ie, `--vnc` and `--vnc-experimental` shouldn't _set_ `noGraphics`, because we're still running with graphics (non-headless), but we are using a different way to provide the graphics, so we skip the call to `runUI`.

Sources/tart/Commands/Run.swift
```diff
@@ -332,8 +332,7 @@ struct Run: AsyncParsableCommand {
     }
     sigusr1Src.activate()
 
-    let useVNCWithoutGraphics = (vnc || vncExperimental) && !graphics
-    if noGraphics || useVNCWithoutGraphics {
+    if noGraphics {
```
torarnv, line 335: Ie, turn this into `if noGraphics || vnc || vncExperimental {`

torarnv:
Ie, right now using `--vnc-experimental` I have to choose between:

1. Getting two UI windows for each VM, one for the built in UI and one for Screen Sharing
1. Passing `--no-graphics` and losing out on Tart automatically opening Screen Sharing with the right address and password

Restoring the previous behavior allows the choice of headless or not (`--no-graphics`), to be independent of _how_ the graphics are provided (default vs `--vnc` vs `--vnc-experimental`).

torarnv:
Proposed fix in https://github.com/cirruslabs/tart/pull/758

PR #735, opened 2024-02-19
Introduce "tart fqn" command

edigaryev:
This allows the `tart pull`'s callee to figure out a fully qualified OCI image name of the pulled image (even if it was already cached).

### DIFF
--- a/Sources/tart/Logging/Logger.swift
+++ b/Sources/tart/Logging/Logger.swift
@@ -12,12 +12,15 @@ var defaultLogger: Logger = {
     return InteractiveConsoleLogger()
   }
 }()
-var jsonLogger: Logger = NopLogger()
 
 public class InteractiveConsoleLogger: Logger {
   private let eraseCursorDown = "\u{001B}[J" // clear entire line
   private let moveUp = "\u{001B}[1A" // move one line up
   private let moveBeginningOfLine = "\r" // 
+
+  public init() {
+
+  }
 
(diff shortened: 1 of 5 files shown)


Sources/tart/Logging/Logger.swift
```diff
@@ -12,16 +12,13 @@ var defaultLogger: Logger = {
     return InteractiveConsoleLogger()
   }
 }()
+var jsonLogger: Logger = NopLogger()
```
fkorotkov, line 15: Why it's a `NopLogger`?

fkorotkov (commented):
What do you think of a separate `tart fqn` hidden (`shouldDisplay` equal to `false`) subcommand that we can run similar to `tart ip`? This can return this fully qualified name of a remote image in the local cache.

Sources/tart/Logging/Logger.swift
edigaryev: Because we don't output JSON by default.

edigaryev:
> What do you think of a separate `tart fqn` hidden (`shouldDisplay` equal to `false`) subcommand that we can run similar to `tart ip`? This can return this fully qualified name of a remote image in the local cache.

See https://github.com/cirruslabs/tart/pull/735/commits/50cdd9546b0753123433e286810dce8c29afdd8f.

fkorotkov (approved):

PR #746, opened 2024-03-08
Improve macOS app integration

torarnv:
Tart is now a proper application bundle, with the name and icon declared in the Info.plist, which we were missing.

This also allows us to declare the app as LSBackgroundOnly as a default, which means that 'tart create' and similar background commands will not show the application icon in the dock, while 'tart run' will, thanks to it overriding the activation policy of the app.

For now the logic of creating the Tart.app bundle is duplicated between the CI packaging scripts and the run-signed.sh script. Now that these scripts are growing, it makes sense to look at whether we can share the logic somehow, e.g. by building the application bundle directly during build, and packaging that, instead of creating it as a post install step.


torarnv:
Proper branding in the UI:

image
image

No change to command line assumptions about the name:

image

edigaryev (approved):

fkorotkov (approved):

torarnv:
Hmm, looks like this actually _regressed_ our app integration. The homebrew formula apparently doesn't install the Info.plist or app icon:

```
❯ find /opt/homebrew/Cellar/tart/2.7.0
/opt/homebrew/Cellar/tart/2.7.0
/opt/homebrew/Cellar/tart/2.7.0/INSTALL_RECEIPT.json
/opt/homebrew/Cellar/tart/2.7.0/LICENSE
/opt/homebrew/Cellar/tart/2.7.0/bin
/opt/homebrew/Cellar/tart/2.7.0/bin/tart
/opt/homebrew/Cellar/tart/2.7.0/.brew
/opt/homebrew/Cellar/tart/2.7.0/.brew/tart.rb
/opt/homebrew/Cellar/tart/2.7.0/libexec
/opt/homebrew/Cellar/tart/2.7.0/libexec/tart.app
/opt/homebrew/Cellar/tart/2.7.0/libexec/tart.app/Contents
/opt/homebrew/Cellar/tart/2.7.0/libexec/tart.app/Contents/MacOS
/opt/homebrew/Cellar/tart/2.7.0/libexec/tart.app/Contents/MacOS/tart
/opt/homebrew/Cellar/tart/2.7.0/libexec/tart.app/Contents/embedded.provisionprofile
```

@fkorotkov @edigaryev What is https://github.com/cirruslabs/tart/blob/main/.ci/create-pkg.sh used for, and what is the relationship to the homebrew packaging? My assumption was that changing https://github.com/cirruslabs/tart/blob/main/.ci/create-pkg.sh and https://github.com/cirruslabs/tart/blob/main/scripts/run-signed.sh was enough to cover both development and production scenarios. 

Without the Info.plist and app icon https://github.com/cirruslabs/tart/pull/746/files#diff-b0065f20172c946df326dee83abae39247b506400501d85c82272ef63211daa1 results in now always bringing up the Dock icon, even for headless runs, and the Dock icon is always missing Tart branded icon 😢 

I'm going to clean up and fix this, sorry for the fallout. 

My initial plan here is to:

 * Ensure that builds signed .app bundle (signed by developer cert or production cert)
 * Set up the GitHub release pipeline to package the app bundle
   * Right now those are reported as broken by macOS, even before this change (https://github.com/cirruslabs/tart/releases/download/2.6.1/tart-arm64.tar.gz e.g.)
 * Build the .ci/ packaging and the homebrew packaging on top of the GitHub released package
   * With a forwarding command line script

Thoughts?

torarnv:
Ah, and I see you already fixed the main issue in https://github.com/cirruslabs/tart/pull/756 . Thank you!

PR #751, opened 2024-03-11
Lowercase `tart.app`

fkorotkov:
So signing and `embedded.provisionprofile` work as expected.

Related to #746

### DIFF
--- a/.ci/pkg/scripts/postinstall
+++ b/.ci/pkg/scripts/postinstall
@@ -3,13 +3,13 @@
 set -e
 
 # fix structure
-mkdir -p "$2/Tart.app/Contents/MacOS" "$2/Tart.app/Resources"
-mv "$2/tart" "$2/Tart.app/Contents/MacOS/tart"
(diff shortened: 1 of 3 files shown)


edigaryev (approved):

fkorotkov:
@torarnv I think we can't rename to `Tart.app` that easily since for `embedded.provisionprofile` to work bianry name and `*.app` package name should be the same including case. We had a lot of issues with singing and other things before. I thinking keeping `tart.app` for now is acceptable. What do you think?

torarnv:
Ah, sorry, wasn't aware of that. Yeah, let's rename back, and I can have a look at whether it's possible to rename while still having signing and everything else working.

torarnv:
Again, sorry for regressing that part 😢

fkorotkov:
No worries. I had a gut feeling that it might be the issue and checked a dry run before releasing a new version of Tart. So no regression was actually introduced to the end users.

PR #753, opened 2024-03-11
tart run: introduce --net-softnet-allow command-line argument

edigaryev:
See https://github.com/cirruslabs/softnet/pull/34.

### DIFF
--- a/Sources/tart/Commands/Run.swift
+++ b/Sources/tart/Commands/Run.swift
@@ -117,6 +117,9 @@ struct Run: AsyncParsableCommand {
                            discussion: "Learn how to configure Softnet for use with Tart here: https://github.com/cirruslabs/softnet"))
   var netSoftnet: Bool = false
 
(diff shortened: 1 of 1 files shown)


Sources/tart/Commands/Run.swift
```diff
@@ -117,6 +117,9 @@ struct Run: AsyncParsableCommand {
                            discussion: "Learn how to configure Softnet for use with Tart here: https://github.com/cirruslabs/softnet"))
   var netSoftnet: Bool = false
 
+  @Option(help: ArgumentHelp("Comma-separated list of CIDRs to allow the traffic to when using Softnet isolation\n(e.g. --net-softnet-allow=192.168.0.0/24)", valueName: "comma-separated CIDRs"))
+  var netSoftnetAllow: String?
```
fkorotkov, line 121: Should it be `[String]` so one can pass multiple `--net-softnet-allow`?

edigaryev: I think we could start with limiting it to just a single argument as it's already tedious to type it multiple times.

fkorotkov (approved):

PR #755, opened 2024-03-12
Fixed `--no-graphics` mode

fkorotkov:
Regression introduced in #746


fkorotkov:
FYI @torarnv. Not sure why `NSApp` is `nil`.

torarnv:
`NSApp` should normally be just a wrapper around `NSApplication.shared`, but I believe _creation_ of the app only happens on explicit invocation, so using `NSApplication.shared` explicitly makes sense.

PR #768, opened 2024-03-26
Fetch IPSW files via Fetch.fetchViaMemory so we can report progress

### LINKED ISSUE #767: Progress during IPSW download
Hi,

I noticed that `tart create --from-ipsw <url>` didn't report progress for the download, and naively implemented:

```diff
diff --cc Sources/tart/Fetcher.swift
index 461ad59,461ad59..ab8a8ee
--- a/Sources/tart/Fetcher.swift
+++ b/Sources/tart/Fetcher.swift
@@@ -29,18 -29,18 +29,26 @@@ class Fetcher 
    private static func fetchViaFile(_ request: URLRequest) async throws -> (AsyncThrowingChannel<Data, Error>, HTTPURLResponse) {
      let dataCh = AsyncThrowingChannel<Data, Error>()
  
--    let (fileURL, response) = try await urlSession.download(for: request)
++    let (asyncBytes, response) = try await urlSession.bytes(for: request)
  
--    // Acquire a handle to the downloaded file and then remove it.
--    //
--    // This keeps a working reference to that file, yet we don't
--    // have to deal with the cleanup any more.
--    let mappedFile = try Data(contentsOf: fileURL, options: [.alwaysMapped])
--    try FileManager.default.removeItem(at: fileURL)
++    let chunkSize = 64 * 1024 * 1024
  
      Task {
--      for chunk in (0 ..< mappedFile.count).chunks(ofCount: 64 * 1024 * 1024) {
--        await dataCh.send(mappedFile.subdata(in: chunk))
++      let expectedLength = response.expectedContentLength
++
++      var chunk = Data()
++      if expectedLength > 0 {
++          chunk.reserveCapacity(min(chunkSize, Int(expectedLength)))
++      } else {
++          chunk.reserveCapacity(chunkSize)
++      }
++
++      for try await byte in asyncBytes {
++        chunk.append(byte)
++        if chunk.count >= chunkSize {
++          await dataCh.send(chunk)
++          chunk.removeAll(keepingCapacity: true)
++        }
        }
  
        dataCh.finish()
```

But looking at `git blame` for this logic there seems to have been quite some variants of the download code (#260, #262, #282, #306), including one that used `bytes`. 

@edigaryev @fiedukow Could you shed some light on the history of `Fetcher.fetchViaFile` and why it is needed? It seems to have been architected in https://github.com/cirruslabs/tart/pull/280#issuecomment-1300105314

For IPSWs, do we gain anything by downloading to a file first (and waiting for it to complete the download), and then delivering that file in chunks to `retrieveIPSW` that writes it to another file (and computes a digest)?

Should `retrieveIPSW` perhaps go via `fetchViaMemory` instead, and `fetchViaMemory` thought to use `bytes` instead of the blocking `data` call?

torarnv:
We're already writing the bytes to a file in the end, so we don't need to go via Fetcher.fetchViaFile. Doing so would also result in waiting for the entire download to complete before reporting any progress.

Fixes #767


Sources/tart/Fetcher.swift
```diff
@@ -15,12 +15,28 @@ class Fetcher {
...
 
-      dataCh.finish()
+      var chunk = Data()
+      if expectedLength > 0 {
+          chunk.reserveCapacity(min(chunkSize, Int(expectedLength)))
+      } else {
+          chunk.reserveCapacity(chunkSize)
+      }
+
+      for try await byte in asyncBytes {
```
edigaryev, line 33: We've used this approach in the past, unfortunately, the `URLSession.bytes()` literally processes each byte individually without any buffering, which results in severe performance degradation, see https://github.com/cirruslabs/tart/pull/282.

torarnv: Hehe, yepp, that's what I saw now when testing localhost. It wasn't as visible when testing against the Apple content servers.

Would restoring the delegate work, or is there no way to have progress and speed at the same time?

edigaryev: >Would restoring the delegate work, or is there no way to have progress and speed at the same time?

/cc @fkorotkov

torarnv: I think I've found a way to make this work, stay tuned :)

Sources/tart/VM.swift
```diff
@@ -91,12 +91,17 @@ class VM: NSObject, VZVirtualMachineDelegate, ObservableObject {
...
-    let progress = Progress(totalUnitCount: response.expectedContentLength)
-    ProgressObserver(progress).log(defaultLogger)
+    let request = URLRequest(url: remoteURL)
+    let (channel, response) = try await Fetcher.fetch(request, viaFile: true, progress: downloadProgress)
 
     let temporaryLocation = try Config().tartTmpDir.appendingPathComponent(UUID().uuidString + ".ipsw")
+    defaultLogger.appendNewLine("Computing digest for \(temporaryLocation.path)...")
+    let digestProgress = Progress(totalUnitCount: response.expectedContentLength)
+    ProgressObserver(digestProgress).log(defaultLogger)
+
```
torarnv, line 104: Why do compute the digest here? We already have the hash from `x-amz-meta-digest-sha256` above. If we don't trust that the download actually delivers a file with the same hash we should probably compare the hashes, but I don't think we need to be that paranoid do we? 

Copying the data via the loop below takes a bit of time, compared to just moving the file, so perhaps we can just move the file from the `URLSession` temporary location into the Tart IPSW cache?

edigaryev: > Why do compute the digest here? We already have the hash from `x-amz-meta-digest-sha256` above.

Because `--from-ipsw` might point to an arbitrary URL, which will not always be an S3 storage.

torarnv: Ah, of course. Since the typical use-case is pulling from Apple's S3 storage, would it make sense to skip the digest for those cases we do have a pre-computed hash and just move the file directly? Would speed up downloads by 30ish seconds in my local testing (even the copy takes some time, even without the digest).

torarnv: Anyways, this is something to consider beyond this PR, so I'll close this conversation as "resolved" to mark this PR as "complete" :)

Sources/tart/Fetcher.swift
```diff
@@ -3,19 +3,37 @@ import AsyncAlgorithms
...
+    self.progress = progress
+  }
+
+  var observation: NSKeyValueObservation?
+  func urlSession(_ session: URLSession, didCreateTask task: URLSessionTask) {
+    observation = task.progress.observe(\.fractionCompleted) { progress, value in
+      self.progress.completedUnitCount = Int64(
+        progress.fractionCompleted * Double(self.progress.totalUnitCount)
+      )
+    }
```
edigaryev, line 18: Looks like it's possible to simplify this down to:

```suggestion
  func urlSession(_ session: URLSession, didCreateTask task: URLSessionTask) {
    self.progress.addChild(task.progress, withPendingUnitCount: 100)
```

torarnv: I tested that initially, but the progress of the parent progress didn't update. That's also in line with the docs at https://developer.apple.com/documentation/foundation/progress?language=objc#1661050

> The [completedUnitCount](https://developer.apple.com/documentation/foundation/progress/1407934-completedunitcount?language=objc) property for a containing progress object only updates when the suboperation is 100% complete. The [fractionCompleted](https://developer.apple.com/documentation/foundation/progress/1408579-fractioncompleted?language=objc) property for a containing progress object updates continuously as work progresses for all suboperations.

I can retry just in case I missed something in my original testing.

edigaryev: >I tested that initially, but the progress of the parent progress didn't update.

Could've been because the `let downloadProgress = Progress()` was not initialized with `totalUnitCount: 100`.

> The [completedUnitCount](https://developer.apple.com/documentation/foundation/progress/1407934-completedunitcount?language=objc) property for a containing progress object only updates when the suboperation is 100% complete. The [fractionCompleted](https://developer.apple.com/documentation/foundation/progress/1408579-fractioncompleted?language=objc) property for a containing progress object updates continuously as work progresses for all suboperations.

I think this is fine because we only use the `fractionCompleted` to report the progress, see `ProgressObserver.lineToRender()`.

torarnv: You're right, thanks!

I pushed an update where I used `withPendingUnitCount: self.progress.totalUnitCount`, so that we reflect the input `Progress` fully regardless of its total count.

edigaryev (approved):

fkorotkov (approved):
Thank you for figuring it out! Me and @edigaryev spent endless hours figuring it out.

Issue #762, opened 2024-03-19
Automate benchmarking


fkorotkov:
I think it's time to automate benchmarking of guest vs host. I propose to create a set of tests that we can benchmark agains. Having it in a separate repository will allow to write it in a nicer Go language.

I think we can start by just testing IO of writing and reading files from disk. Test will determine host macOS version, pull corresponding vanilla image, and run the same command on host vs guest. Will be great to compare 1Mb, 10Mb, 100Mb, 1000Mb random file writes.

paragchandra:
Hey, just happened upon this issue while I was checking on the status of a bug I filed yesterday (which you guys have already fixed; thanks for that!). A while back (summer 2023) I used `fio` to measure disk performance inside a Tart VM, outside the VM, and using the shared file system. I'm paraphrasing from the internal write-up I did below; could be an option to consider for benchmarking:

[fio](https://github.com/axboe/fio). This utility has many options for testing all manner of different scenarios, but to keep things simple I followed the [guide at Ars Technica](https://arstechnica.com/gadgets/2020/02/how-fast-are-your-disks-find-out-the-open-source-way-with-fio/) to get a sense of best-case, worst-case, and average-case performance of three different filesystems:

* Inside VM - the tests are performed inside the VM, on the VM’s own filesystem;
* Outside VM - directly on the machine which hosts the VM;
* Shared FS - make use of Tart’s shared folder feature to mount a directory from the host directly into a folder inside the VM, and run the tests in this shared filesystem

Scenario | Worst Case | Average Case | Best Case
-- | -- | -- | --
Inside VM | 9826 kB/s | 21.0 GB/s | 377 MB/s
Outside VM | 18.3 MB/s | 24.2 GB/s | 1262 MB/s
Shared FS | 20.4 MB/s | 2351 MB/s* | 1178 MB/s

*I re-ran this just to check, and yes, this scenario was an order of magnitude slower

fkorotkov:
@paragchandra thanks for sharing! `fio` looks interesting and seems like a good tool to use!

Can I ask you why Best Case is slower than Average Case in your table? Otherwise average looks as I remember measuring manually: about 10-15% slower.

PR #769, opened 2024-03-26
Introduce Golang-based benchmarking utility

edigaryev:
Resolves https://github.com/cirruslabs/tart/issues/762.


benchmark/README.md
```diff
@@ -0,0 +1,15 @@
+# Benchmark
```
fkorotkov, line 1: Maybe run it on the M2 Pro `dev-mini` host we have a include the run information at the end of readme? Including host and guest macOS versions (should be 14.4.1).

edigaryev: Added in https://github.com/cirruslabs/tart/pull/769/commits/f0adb8b016a09eba67f185ae639888683824e223.

benchmark/internal/executor/tart/tart.go
```diff
@@ -0,0 +1,133 @@
...
+
+	if err := Cmd(ctx, tart.logger, "clone", baseImage, tart.vmName); err != nil {
+		return nil, err
+	}
+
+	vmRunCtx, vmRunCancel := context.WithCancel(ctx)
+	tart.vmRunCancel = vmRunCancel
+
+	go func() {
+		_ = Cmd(vmRunCtx, tart.logger, "run", "--no-graphics", tart.vmName)
```
fkorotkov, line 44: Do you think it makes sence to match guest CPU/memory with the host. I guess for the current IO testing it won't make a difference but in the future we'll need to have such arguments.

edigaryev: I think this totally depends on what we're trying to benchmark for. For now the defaults look reasonable to me.

fkorotkov (approved):

PR #786, opened 2024-04-11
Allow mounting NBD disks

### LINKED ISSUE #759: Enable the possibility to mount NBD devices
Hi,
I would like to mount a physical disk from a NAS using NBD as main disk for a VM.

I suppose this should be possible as there is
VZNetworkBlockDeviceStorageDeviceAttachment -> https://developer.apple.com/documentation/virtualization/vznetworkblockdevicestoragedeviceattachment

Ideally it would be nice to provide a nbd url as disk so it would be able to mount on start.

Thanks for everything

fkorotkov:
Fixes #759

So people can experiment with mounting NBDs before we can decide if it's viable to put the whole VM on it.

### DIFF
--- a/Sources/tart/Commands/Run.swift
+++ b/Sources/tart/Commands/Run.swift
@@ -438,10 +438,10 @@ struct Run: AsyncParsableCommand {
           throw RuntimeError.VMConfigurationError("invalid NBD URL: \(diskPath)")
         }
         let nbdAttachment = try VZNetworkBlockDeviceStorageDeviceAttachment(
-            url: nbdURL,
-            timeout: 30,
-            isForcedReadOnly: diskReadOnly,
-            synchronizationMode: VZDiskSynchronizationMode.none
+          url: nbdURL,
+          timeout: 30,
+          isForcedReadOnly: diskReadOnly,
+          synchronizationMode: VZDiskSynchronizationMode.none
         )
         result.append(VZVirtioBlockDeviceConfiguration(attachment: nbdAttachment))
         continue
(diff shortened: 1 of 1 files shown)


Sources/tart/Commands/Run.swift
```diff
@@ -429,6 +431,25 @@ struct Run: AsyncParsableCommand {
...
       let diskPath = diskReadOnly ? String(rawDisk.prefix(rawDisk.count - readOnlySuffix.count)) : rawDisk
+
+      if (diskPath.starts(with: "nbd://")) {
+        guard #available(macOS 14, *) else {
+          throw UnsupportedOSError("attaching Network Block Devices", "are")
+        }
+        let nbdURL =  URL(string: diskPath)
+        if nbdURL == nil {
+          throw RuntimeError.VMConfigurationError("invalid NBD URL: \(diskPath)")
+        }
```
edigaryev, line 442:
```suggestion
        guard let nbdURL = URL(string: diskPath) else {
          throw RuntimeError.VMConfigurationError("invalid NBD URL: \(diskPath)")
        }
```

Sources/tart/Commands/Run.swift
```diff
@@ -429,6 +431,25 @@ struct Run: AsyncParsableCommand {
...
+        let nbdAttachment = try VZNetworkBlockDeviceStorageDeviceAttachment(
+            url: nbdURL!,
```
edigaryev, line 444:
```suggestion
            url: nbdURL,
```

Sources/tart/Commands/Run.swift
```diff
@@ -72,6 +72,8 @@ struct Run: AsyncParsableCommand {
   To workaround this issue pass TART_HOME explicitly:
 
   sudo TART_HOME="$HOME/.tart" tart run sonoma --disk=/dev/disk0
+
+  To work with NBD devices use the following syntax:
```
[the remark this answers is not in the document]
fkorotkov: Removed it in 9ee5247ca6b67684daacbcfe1f273bb8ef09e681 since there is an example in the first line.

edigaryev (approved):

PR #787, opened 2024-04-11
Fixed Swift Warning

### LINKED ISSUE #785: Swift compiler warnings
When building I noticed the compilers is being a bit angry at some stuff:

```
$ ./scripts/run-signed.sh
Building for debugging...
/Users/bartek/projects/tart/Sources/tart/OCI/Registry.swift:45:7: warning: immutable property will not be decoded because it is declared with an initial value which cannot be overwritten
  let defaultIssuedAt = Date()
      ^
/Users/bartek/projects/tart/Sources/tart/OCI/Registry.swift:45:7: note: set the initial value via the initializer or explicitly define a CodingKeys enum without a 'defaultIssuedAt' case to silence this warning
  let defaultIssuedAt = Date()
      ^
/Users/bartek/projects/tart/Sources/tart/OCI/Registry.swift:45:7: note: make the property mutable instead
  let defaultIssuedAt = Date()
  ~~~ ^
  var
/Users/bartek/projects/tart/Sources/tart/OCI/Registry.swift:46:7: warning: immutable property will not be decoded because it is declared with an initial value which cannot be overwritten
  let defaultExpiresIn = 60
      ^
/Users/bartek/projects/tart/Sources/tart/OCI/Registry.swift:46:7: note: set the initial value via the initializer or explicitly define a CodingKeys enum without a 'defaultExpiresIn' case to silence this warning
  let defaultExpiresIn = 60
      ^
/Users/bartek/projects/tart/Sources/tart/OCI/Registry.swift:46:7: note: make the property mutable instead
  let defaultExpiresIn = 60
  ~~~ ^
  var
/Users/bartek/projects/tart/Sources/tart/Commands/Suspend.swift:17:9: warning: variable 'pid' was never mutated; consider changing to 'let' constant
    var pid = try lock.pid()
    ~~~ ^
    let
/Users/bartek/projects/tart/Sources/tart/Commands/Suspend.swift:17:9: warning: variable 'pid' was never mutated; consider changing to 'let' constant
    var pid = try lock.pid()
    ~~~ ^
    let
/Users/bartek/projects/tart/Sources/tart/VM.swift:249:12: warning: instance method 'wait' is unavailable from asynchronous contexts; Await a Task handle instead; this is an error in Swift 6
      sema.wait()
           ^
Dispatch.DispatchSemaphore:4:17: note: 'wait()' declared here
    public func wait()
                ^
/Users/bartek/projects/tart/Sources/tart/VM.swift:249:12: warning: instance method 'wait' is unavailable from asynchronous contexts; Await a Task handle instead; this is an error in Swift 6
      sema.wait()
           ^
Dispatch.DispatchSemaphore:4:17: note: 'wait()' declared here
    public func wait()
                ^
[30/30] Applying tart
Build complete! (3.18s)
```

fkorotkov:
Plus updated all the dependencies and Swift Tools.

Fixes #785

### DIFF
--- a/Tests/TartTests/TokenResponseTests.swift
+++ b/Tests/TartTests/TokenResponseTests.swift
@@ -3,24 +3,26 @@ import XCTest
 
 final class TokenResponseTests: XCTestCase {
   func testBasic() throws {
+    let expectedTokenExpiresAtRange = DateInterval()
     let tokenResponseRaw = Data("{\"token\":\"some token\"}".utf8)
     let tokenResponse = try TokenResponse.parse(fromData: tokenResponseRaw)
 
     XCTAssertEqual(tokenResponse.token, "some token")
 
-    let expectedTokenExpiresAtRange = Date()...Date().addingTimeInterval(60)
+    expectedTokenExpiresAtRange.end = Date().addingTimeInterval(60)
     XCTAssertTrue(expectedTokenExpiresAtRange.contains(tokenResponse.tokenExpiresAt))
 
     XCTAssertTrue(tokenResponse.isValid())
   }
 
   func testExpirationBasic() throws {
+    let expectedTokenExpiresAtRange = DateInterval()
     let tokenResponseRaw = Data("{\"token\":\"some token\",\"expires_in\":2}".utf8)
     let tokenResponse = try TokenResponse.parse(fromData: tokenResponseRaw)
 
     XCTAssertEqual(tokenResponse.expiresIn, 2)
 
-    let expectedTokenExpiresAtRange = Date()...Date().addingTimeInterval(2)
+    expectedTokenExpiresAtRange.end = Date().addingTimeInterval(2)
     XCTAssertTrue(expectedTokenExpiresAtRange.contains(tokenResponse.tokenExpiresAt))
 
     XCTAssertTrue(tokenResponse.isValid())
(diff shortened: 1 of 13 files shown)


edigaryev (approved):

Tests/TartTests/TokenResponseTests.swift
```diff
@@ -3,24 +3,26 @@ import XCTest
...
 final class TokenResponseTests: XCTestCase {
   func testBasic() throws {
+    var expectedTokenExpiresAtRange = DateInterval()
     let tokenResponseRaw = Data("{\"token\":\"some token\"}".utf8)
     let tokenResponse = try TokenResponse.parse(fromData: tokenResponseRaw)
 
     XCTAssertEqual(tokenResponse.token, "some token")
 
-    let expectedTokenExpiresAtRange = Date()...Date().addingTimeInterval(60)
+    expectedTokenExpiresAtRange.end = Date().addingTimeInterval(60)
```
edigaryev, line 12: Do we need to change this file at all?

fkorotkov: I think so. I saw it flaking before and I think there is a race in the start of the interval. When we might get a newer start and than the tests fails if we are not running it within the same millisecond.

edigaryev (approved):

Issue #791, opened 2024-04-15
documentation: how to mount a virtiofs (--dir) folder  using just /etc/fstab. 


tschiex:
`mount -t virtiofs com.apple.virtio-fs.automount /mnt/shared
`
works. 

The documentation could explain, if this is feasible, how I could achieve the same using just `/etc/fstab`,
 assuming eg. I used `--dir=mac:/Users/me/Documents`.

edigaryev:
I don't think that `/etc/fstab` is a thing on newer macOS versions, but perhaps someone more knowledgeable on the topic will chime in.

tschiex:
Oops, I should have mentioned that this is for a Linux VM guest, not a MacOS guest.
For Linux, /etc/fstab is still widely used (even if often managed by systems I think).

PR #793, opened 2024-04-17
Document how to mount the shared directory on Linux at boot time

edigaryev:
Resolves https://github.com/cirruslabs/tart/issues/791.

### DIFF
--- a/docs/quick-start.md
+++ b/docs/quick-start.md
@@ -143,7 +143,14 @@ Note: to use the directory mounting feature, the guest VM needs to run macOS 13.
 To be able to access the shared directories from the Linux guest, you need to manually mount the virtual filesystem first:
 
 ```bash
+mkdir /mnt/shared
 mount -t virtiofs com.apple.virtio-fs.automount /mnt/shared
 ```
 
 The directory we've mounted above will be accessible from the `/mnt/shared/project` path inside a guest VM.
+
+To automatically mount this directory at boot time, add the following line to the `/etc/fstab` file:
+
+```shell
+com.apple.virtio-fs.automount /mnt/shared virtiofs rw,relatime 0 0
+```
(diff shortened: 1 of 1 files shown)


docs/quick-start.md
```diff
@@ -143,7 +143,14 @@ Note: to use the directory mounting feature, the guest VM needs to run macOS 13.
...
 mount -t virtiofs com.apple.virtio-fs.automount /mnt/shared
 ```
 
 The directory we've mounted above will be accessible from the `/mnt/shared/project` path inside a guest VM.
+
+To automatically mount this directory at boot time, add the following line to the `/etc/fstab` file:
+
+```shell
+com.apple.virtio-fs.automount /mnt/shared virtiofs rw,relatime 0 0
+```
```
fkorotkov, line 156: Should we make it an expandable [admonition](https://squidfunk.github.io/mkdocs-material/reference/admonitions/) so it only takes one line?

```suggestion
??? info "Auto-mount at boot time"
    To automatically mount this directory at boot time, add the following line to the `/etc/fstab` file:

    ```shell
    com.apple.virtio-fs.automount /mnt/shared virtiofs rw,relatime 0 0
    ```
```

tschiex:
Thanks. 

Currently, every file of the mounted APFS folder, as a virtiofs volume, appears as belonging to the root user in the Linux guest.

Would it be possible to preserve the APFS uid of every file in the Linux guest?

edigaryev:
>Would it be possible to preserve the APFS uid of every file in the Linux guest?

I'm not sure if this is possible, but you can `chown <USER>:<GROUP> /mnt/shared`, though, and then add the `user` option to the `/etc/fstab` entry:

```diff
-com.apple.virtio-fs.automount /mnt/shared virtiofs rw,relatime 0 0
+com.apple.virtio-fs.automount /mnt/shared virtiofs rw,relatime,user 0 0
```

This will change the ownership of files in `/mnt/shared` to `<USER>:<GROUP`.

On macOS, `mount_virtiofs` has `-u` and `-g` flags for that.

PR #801, opened 2024-04-23
Check all VMs for MAC collision

fkorotkov:
Before only suspendable VMs were getting checked. Not sure why. It makes sense to check all.

### DIFF
--- a/Sources/tart/Commands/Run.swift
+++ b/Sources/tart/Commands/Run.swift
@@ -187,15 +187,19 @@ struct Run: AsyncParsableCommand {
     let localStorage = VMStorageLocal()
     let vmDir = try localStorage.open(name)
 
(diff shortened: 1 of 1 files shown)


edigaryev:
>Before only suspendable VMs were getting checked. Not sure why.

Because for the rest of the VMs we do this in `tart clone`:

https://github.com/cirruslabs/tart/blob/9f38441a42dcfe6180c5f0be4e9a7c42c6755338/Sources/tart/Commands/Clone.swift#L65-L67

fkorotkov:
There was a report when a person just `scp`ed the same VM twice. 🤷  Seems like an improvement.

edigaryev:
> There was a report when a person just `scp`ed the same VM twice. 🤷 Seems like an improvement.

Let's acquire an unconditional global lock then to simplify things up.

edigaryev (approved):

PR #802, opened 2024-04-23
Separate tart engine and commands into separate modules

krzyzanowskim:
An attempt to separate code into two targets
- **tart** commands
- **TartEngine**

my goal is to make tart logic available as a **Tart** library. This is the first attempt that separate code, and the following step is to make the public API. For now, commands rely on the TartEngine API.

I'm aware this is a significant change to the project, nobody asked. I put it here to start a discussion about whether this is a direction maintainers are willing to move the project forward.


fkorotkov:
Hey @krzyzanowskim, thank you for working on this! We have some very distant plans to have a standalone Desktop application #305 at some point. There are some internal ideas for the desktop app too to extend beyond just running VMs. Don't have engineering resources to work on this just yet though.

In any case this work seems to be useful for our eventual goal. I'm wondering what is your motivation here? Are you planning to depend on TartEngine?

fkorotkov:
Closing for now. We might work on something like that in the future but have no plans at the moment.

PR #806, opened 2024-04-26
Add support pasting clipboard from host for Linux VMs

amalchuk:
Partially closing #14.

Tested on Ubuntu 22.04, and works only on Linux VMs only for now.

### DIFF
--- a/Sources/tart/Commands/Run.swift
+++ b/Sources/tart/Commands/Run.swift
@@ -40,6 +40,11 @@ struct Run: AsyncParsableCommand {
   @Flag(help: "Disable audio pass-through to host.")
   var noAudio: Bool = false
 
+  @Flag(help: ArgumentHelp(
+    "Disable clipboard sharing between host and guest.",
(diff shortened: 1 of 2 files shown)


edigaryev:
Hello Andrew, thanks for the contribution!

Would you mind addressing the concerns expressed https://github.com/cirruslabs/tart/pull/202#issuecomment-1228936229 in this PR too?

Or we can merge and then build add an option on top of this.

/cc @fkorotkov

amalchuk:
What do you think about creating an option `"audio": bool` in config.json for this feature?
Or, yet another option in CLI – such as `--no-audio`, create an option `--no-clipboard`?

For now, every features is enabled by default – audio is one, for example. If the user doesn't want to use one of the default features, we can provide the options for disable them.

I think, enabling clipboard sharing by default is good idea, and creating the option `--no-clipboard` for disable this feature too. Agree?

fkorotkov (approved):

edigaryev (approved):

PR #808, opened 2024-04-29
Gracefully stop vm on tart stop

torarnv:
We were letting the CancellationError bubble up all the way until
it terminated app, which meant we didn't hit the shutdown code
in run(), stopping the VM and the network.

We now catch CancellationError and proceed to gracefully shut down.

We only stop the VM if it's still running, as a VM that has been
stopped via the menu can't be stopped again.

### DIFF
--- a/Sources/tart/Commands/Run.swift
+++ b/Sources/tart/Commands/Run.swift
@@ -567,69 +567,69 @@ struct Run: AsyncParsableCommand {
     nsApp.setActivationPolicy(.regular)
     nsApp.activate(ignoringOtherApps: true)
 
-    struct MainApp: App {
-      static var suspendable: Bool = false
-      static var capturesSystemKeys: Bool = false
+    MainApp.suspendable = suspendable
+    MainApp.capturesSystemKeys = captureSystemKeys
+    MainApp.main()
+  }
+}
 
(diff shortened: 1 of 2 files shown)


Sources/tart/Commands/Run.swift
```diff
@@ -563,73 +563,70 @@ struct Run: AsyncParsableCommand {
...
+      // Remove some standard menu options
+      CommandGroup(replacing: .help, addition: {})
+      CommandGroup(replacing: .newItem, addition: {})
+      CommandGroup(replacing: .pasteboard, addition: {})
+      CommandGroup(replacing: .textEditing, addition: {})
+      CommandGroup(replacing: .undoRedo, addition: {})
+      CommandGroup(replacing: .windowSize, addition: {})
+      // Replace some standard menu options
+      CommandGroup(replacing: .appInfo) { AboutTart(config: vm!.config) }
+      //CommandGroup(replacing: .appTermination) { AboutTart(config: vm!.config) }
```
edigaryev, line 610:
```suggestion
```

torarnv: Doh! Leftover from testing :)

edigaryev (approved):

fkorotkov (approved):

PR #809, opened 2024-05-01
Introduce `--random-mac` and `--random-serial` flags for `tart set`

fkorotkov:
To generate new MAC address and/or serial number for a given VM.

### DIFF
--- a/Sources/tart/Commands/Set.swift
+++ b/Sources/tart/Commands/Set.swift
@@ -1,5 +1,6 @@
 import ArgumentParser
 import Foundation
+import Virtualization
 
 struct Set: AsyncParsableCommand {
   static var configuration = CommandConfiguration(commandName: "set", abstract: "Modify VM's configuration")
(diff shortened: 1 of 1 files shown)


Sources/tart/Commands/Set.swift
```diff
@@ -51,6 +60,17 @@ struct Set: AsyncParsableCommand {
       }
     }
 
+    if randomMAC {
+      vmConfig.macAddress = VZMACAddress.randomLocallyAdministered()
+    }
+
+    #if arch(arm64)
+      if randomSerial {
+        let oldPlatform = vmConfig.platform as! Darwin
```
edigaryev, line 69: Should we not assume that it's always Darwin and do a proper type check & throw an error if it's Linux?

fkorotkov: Though about it but was too lazy and the error message will be self-explanatory. Plus help mentions macOS.

edigaryev (approved):

PR #821, opened 2024-05-14
tart push: use fixed size chunks to allow for better deduplication

edigaryev:
With this PR, we'll yield exactly 2 layers per GB of VM disk image when doing `tart push`, which at the moment equates to:

* 100 layers for `ghcr.io/cirruslabs/macos-sonoma-base:latest` instead of the current 38 layers
* 256 layers for `ghcr.io/cirruslabs/macos-runner:sonoma` instead of the current 162 layers

This slightly increases the bookkeeping costs, but allows for better deduplication in `tart pull` (in the upcoming PR).

Our calculations have shown that current layer deduplication ratios are:

* 0% for distant and 63% for nearby tags for `ghcr.io/cirruslabs/macos-sonoma-base:latest` images
* 0% for distant and 14% for nearby tags for `ghcr.io/cirruslabs/macos-runner:sonoma` images

With the fixed-size chunks of 500 MB, the deduplication ratios will be as follows:

* 46% for distant and 73% for nearby tags for `ghcr.io/cirruslabs/macos-sonoma-base:latest` images
* 36% for distant and 36% for nearby tags for `ghcr.io/cirruslabs/macos-runner:sonoma` images

### DIFF
--- a/Sources/tart/OCI/Layerizer/DiskV2.swift
+++ b/Sources/tart/OCI/Layerizer/DiskV2.swift
@@ -13,12 +13,9 @@ class DiskV2: Disk {
     let mappedDisk = try Data(contentsOf: diskURL, options: [.alwaysMapped])
 
     // Compress the disk file as multiple individually decompressible streams,
-    // each equal ``Self.layerLimitBytes`` bytes or slightly larger due to the
-    // internal compressor's buffer
-    var offset: UInt64 = 0
-
-    while let (compressedData, uncompressedSize, uncompressedDigest) = try compressNextLayerOfLimitBytesOrMore(mappedDisk: mappedDisk, offset: offset) {
-      offset += uncompressedSize
+    // each equal ``Self.layerLimitBytes`` bytes or less due to LZ4 compression
+    for data in mappedDisk.chunks(ofCount: layerLimitBytes) {
+      let compressedData = try (data as NSData).compressed(using: .lz4) as Data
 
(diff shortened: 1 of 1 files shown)


Sources/tart/OCI/Layerizer/DiskV2.swift
```diff
@@ -13,25 +13,22 @@ class DiskV2: Disk {
...
 
     // Compress the disk file as multiple individually decompressible streams,
-    // each equal ``Self.layerLimitBytes`` bytes or slightly larger due to the
-    // internal compressor's buffer
-    var offset: UInt64 = 0
-
-    while let (compressedData, uncompressedSize, uncompressedDigest) = try compressNextLayerOfLimitBytesOrMore(mappedDisk: mappedDisk, offset: offset) {
-      offset += uncompressedSize
+    // each equal ``Self.layerLimitBytes`` bytes or less due to LZ4 compression
+    for data in mappedDisk.chunks(ofCount: layerLimitBytes) {
```
fkorotkov, line 17: Should we put `layerLimitBytes` in the manifest in case it's gonna change at some point?

edigaryev: I think it's already there by the virtue of passing the `UInt64(data.count)` as `uncompressedSize` argument: https://github.com/cirruslabs/tart/blob/13d5ddb4a4d79c0d7babfb24c23436834c1981d2/Sources/tart/OCI/Manifest.swift#L94-L96

fkorotkov (approved):

Issue #827, opened 2024-05-16
Pulling an image from OCI registry with custom port fails to show up in Tart list


robinjohngraham:
My corporate team has provisioned a docker registry on a Nexus server using a custom port of 8221.
I can login and push images to the registry without issue.

But when I run "tart pull registry.fq.d.n:8221:/appleimages/imagename:latest" the image is pulled fully in tact to ~/.tart/cache/OCIs but running "tart list" the image does not show up.

I noticed the image gets stored under ~/.tart/cache/OCIs in a subdirectory that includes the port: (ie. ~/.tart/cache/OCIs/registry.fq.d.n:8221 )

If I rename the folder to just  "~/.tart/cache/OCIs/registry.fq.d.n" and then run "tart list" again, the image shows up in the list.

PR #828, opened 2024-05-17
VMStorageOCI: percent-encode the colon in RemoteName's host

edigaryev:
Resolves https://github.com/cirruslabs/tart/issues/827.

### DIFF
--- a/Sources/tart/VMStorageOCI.swift
+++ b/Sources/tart/VMStorageOCI.swift
@@ -113,7 +113,7 @@ class VMStorageOCI: PrunableStorage {
         continue
       }
 
-      let parts = [foundURL.deletingLastPathComponent().relativePath, foundURL.lastPathComponent]
+      let parts = [foundURL.deletingLastPathComponent().relativePath.percentDecoding(), foundURL.lastPathComponent]
       var name: String
 
       let isSymlink = try foundURL.resourceValues(forKeys: [.isSymbolicLinkKey]).isSymbolicLink!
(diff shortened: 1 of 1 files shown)


fkorotkov:
Should we just ignore port for storing OCI images? Don't think anyone will ever have two registries on different ports. 🤔

edigaryev:
> Should we just ignore port for storing OCI images? Don't think anyone will ever have two registries on different ports. 🤔

IMO we should not assume anything about things that are security-related. It's hard to say exactly what registry configurations people could be using.

fkorotkov:
I just don't like String extensions that are used at specific location. This is an edge case for hosts and preferably this logic should be isolated. For example, I'll need time to understand why [this line](https://github.com/cirruslabs/tart/pull/828/files#diff-52d106b7150fe2a5d777ea88144bbf0a07f23d521a973fa1fd92133ab223fec3R116) is changed and think if we covered all the places. There is a feeling this host specific logic should be better scoped.

edigaryev:
> I just don't like String extensions that are used at specific location.

Switched to an ordinary funcions in https://github.com/cirruslabs/tart/pull/828/commits/09196066274040add041fdaeaa8cf60cbf081294 and added a comment explaining this: https://github.com/cirruslabs/tart/blob/7c386e3466aaa4afe22732b74434a464555d6135/Sources/tart/VMStorageOCI.swift#L116

fkorotkov (approved):

PR #837, opened 2024-06-02
tart run: add --net-softnet-disable-isolation command-line argument

devsunb:
See https://github.com/cirruslabs/softnet/pull/38, https://github.com/cirruslabs/softnet/issues/37

### DIFF
--- a/Sources/tart/Commands/Run.swift
+++ b/Sources/tart/Commands/Run.swift
@@ -127,6 +127,9 @@ struct Run: AsyncParsableCommand {
   @Option(help: ArgumentHelp("Comma-separated list of CIDRs to allow the traffic to when using Softnet isolation\n(e.g. --net-softnet-allow=192.168.0.0/24)", valueName: "comma-separated CIDRs"))
   var netSoftnetAllow: String?
 
+  @Flag(help: ArgumentHelp("Disable network interface isolation to enable communication between VMs when using Softnet isolation"))
+  var netSoftnetDisableIsolation: Bool = false
+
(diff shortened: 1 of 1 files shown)


edigaryev:
Softnet's primary goal is VM network isolation.

Why use it if you don't need that property? 🤔

devsunb:
I haven't found a way to avoid the PRIVATE flag when a VM network interface is added as a member to a bridge network interface while using the default shared NAT network, so I tried using softnet. I don't have a lot of knowledge about this, so I think I may have used the wrong method. Is there any way to use the default shared NAT network and allow VM to VM communication?

edigaryev:
>I haven't found a way to avoid the PRIVATE flag when a VM network interface is added as a member to a bridge network interface while using the default shared NAT network

Interesting. When we were developing Softnet that wasn't the case on macOS 12 (Monterey), but that seems to be changed since macOS Ventura (13)/Sonoma (14).

Have you tried using `--net-bridged` instead? It doesn't have such limitation.

devsunb:
Yes. Using `--net-bridged` works fine, but I need an independent NAT network within my macOS host.

edigaryev:
It feels to me that the naming of `--net-softnet-disable-isolation` is an oxymoron and that it'll make lots of people scratch their head trying to figure out what Softnet even offers in terms of security after all.

What I propose instead is to disable the interface isolation automatically when `--net-softnet-allow=0.0.0.0/0` is passed to Tart.

This way no additional command-line arguments needs to be introduced. What do you think?

edigaryev:
Superseded by https://github.com/cirruslabs/tart/pull/853.

PR #842, opened 2024-06-13
Hook SIGUSR2 to requestStop

torarnv:
For macOS this brings up a dialog, asking the user if they are sure they want to shut down, which makes this less useful for automated graceful shutdowns, but it may behave better on Linux, and there might be ways to instruct macOS to not ask the user, so it's still a nice feature, and aligns with the SIGUSR1 for suspend, and SIGINT for non-graceful shutdown.

### DIFF
--- a/Sources/tart/Commands/Run.swift
+++ b/Sources/tart/Commands/Run.swift
@@ -359,6 +359,18 @@ struct Run: AsyncParsableCommand {
     }
     sigusr1Src.activate()
 
+    // Gracefull shutdown support. For macOS this brings up a dialog,
+    // asking the user if they are sure they want to shut down.
+    signal(SIGUSR2, SIG_IGN)
+    let sigusr2Src = DispatchSource.makeSignalSource(signal: SIGUSR2)
+    sigusr2Src.setEventHandler {
+      Task {
+        print("Requesting guest OS to stop...")
+        try vm!.virtualMachine.requestStop()
+      }
+    }
+    sigusr2Src.activate()
+
     let useVNCWithoutGraphics = (vnc || vncExperimental) && !graphics
     if noGraphics || useVNCWithoutGraphics {
       // enter the main even loop, without bringing up any UI,


Sources/tart/Commands/Run.swift
```diff
@@ -359,6 +359,18 @@ struct Run: AsyncParsableCommand {
...
     sigusr1Src.activate()
 
+    // Gracefull shutdown support. For macOS this brings up a dialog,
+    // asking the user if they are sure they want to shut down.
+    signal(SIGUSR2, SIG_IGN)
+    let sigusr2Src = DispatchSource.makeSignalSource(signal: SIGUSR2)
+    sigusr2Src.setEventHandler {
+      Task {
+        print("Requesting guest OS to stop...")
+        try await vm!.virtualMachine.requestStop()
```
edigaryev, line 369:
```suggestion
        try vm!.virtualMachine.requestStop()
```

edigaryev:
Would you mind elaborating a bit on the concrete use-case you're attempting to cover with this?

I have a rough guess from https://github.com/cirruslabs/packer-plugin-tart/pull/153, but it still would be nice to have an exact reason for the history purposes.

torarnv:
> Would you mind elaborating a bit on the concrete use-case you're attempting to cover with this?
> 
> I have a rough guess from [cirruslabs/packer-plugin-tart#153](https://github.com/cirruslabs/packer-plugin-tart/pull/153), but it still would be nice to have an exact reason for the history purposes.

Sure 🙂 The use-case originated from cirruslabs/packer-plugin-tart#153 as you mention.

When automating Tart (packer plugin, or other CI integrations) you generally want to stop the VM in as clean of a fashion as possible. The packer plugin does this via `shutdown -h now` over SSH, which relies on there being a SSH communicator, and strictly speaking assumes that the guest OS supports that command. 

To compliment that, as a last resort, if all else fails, the VM can be stopped without involving the guest via https://developer.apple.com/documentation/virtualization/vzvirtualmachine/3763076-stop, using `SIGINT`.

This PR adds the ability for automation to ask the VM guest to shut down, via https://developer.apple.com/documentation/virtualization/vzvirtualmachine/3656700-requeststop, which allows a more graceful stop flow. E.g. the automation would first ask the guest for a graceful stop, and if not stopped within 1 minute would then request a hard stop (as implemented in cirruslabs/packer-plugin-tart#153).

It also compliments the suspend, available via `SIGUSR1`.

In practice this feature is of limited usefulness for macOS specifically, due to the confirmation dialog macOS throws up:

Screenshot 2024-06-13 at 13 30 45

But other guest OS'es (Linux) may not have that issue, so exposing it as a general feature of Tart seems reasonable.

edigaryev:
@torarnv thanks for the explanation.

LGTM, there's one nit in https://github.com/cirruslabs/tart/pull/842/files/f792ccfa1205b94222dbfe98dd51cc362c6b283f#diff-b0065f20172c946df326dee83abae39247b506400501d85c82272ef63211daa1, though.

Sources/tart/Commands/Run.swift
torarnv: Done, thanks!

torarnv:
Thanks, fixed nit :)

edigaryev (approved):

PR #843, opened 2024-06-17
Drop Monterey Support

### LINKED ISSUE #841: macOS 15: Enabling VoiceOver crashes VM
Likely a Virtualization.framework issue.

``` 
tart run macos:15.0.0-beta+test
guest has stopped the virtual machine due to error: Error Domain=VZErrorDomain Code=1 "The virtual machine stopped unexpectedly." UserInfo={NSLocalizedFailure=Internal Virtualization error., NSLocalizedFailureReason=The virtual machine stopped unexpectedly.}
```

fkorotkov:
People will still be able to run and SSH into Monterey VMs or use VNC but pointing devices/keyboard won't work.

Fixes #841

### DIFF
--- a/Sources/tart/Commands/Run.swift
+++ b/Sources/tart/Commands/Run.swift
@@ -172,7 +172,7 @@ struct Run: AsyncParsableCommand {
 
     if suspendable {
       let config = try VMConfig.init(fromURL: vmDir.configURL)
-      if !(config.platform is PlatformSuspendable) {
+      if !(config.platform is Darwin) {
         throw ValidationError("You can only suspend macOS VMs")
       }
       if dir.count > 0 {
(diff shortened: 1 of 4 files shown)


edigaryev (dismissed):

edigaryev (approved):

torarnv:
According to the Feedback Assistant report I filed for this the issue has been fixed now in macOS, so we can revert this change I think. I will verify that the issue has been fully fixed.

iby:
I 100% understand how this is a good trade off to keep moving forward. I realize that my use case is not necessarily aligned with your objectives, but ability to run and test against different (especially older) environments is a big part why many of us use VMs in the first place. Regarding CI, in my experience working with small teams, there's no bigger pain than discovering that your CI stack covers 99% of what your requirements, and you need to go and reinvent the wheel for that 1%.

I mostly do macOS development and I've been using Tart more and more lately specifically to maintain older builds. It would be so great to keep older macOS versions compatible and getting support.

torarnv:
The issue that caused this is fixed when running  macOS 15 (beta 4 or above) as host. We can add a runtime check for that and enable the code paths again, making it possible to virtualize macOS 12 with mouse/keyboard on macOS 15.

fkorotkov:
This is reverted in `2.18.0`. 👌

PR #847, opened 2024-06-25
Support customizing VM disks and mounting remote VMs in `tart run`

edigaryev:
To create a VM with a custom disk:

```shell
tart create --linux xcode-16.0
tart set --disk <path to DMG> xcode-16.0
```

To mount this VM after pushing (and pulling it):

```shell
tart run --disk ghcr.io/cirruslabs/xcode:16.0 <VM>
```

### DIFF
--- a/Sources/tart/Commands/Run.swift
+++ b/Sources/tart/Commands/Run.swift
@@ -505,7 +505,7 @@ struct Run: AsyncParsableCommand {
 
         // Unfortunately, VZDiskImageStorageDeviceAttachment does not support
         // FileHandle, so we can't easily clone the disk, open it and unlink(2)
-        // to simplify the garbage collection, so use the <>
+        // to simplify the garbage collection, so use an intermediate directory.
         let clonedDiskURL = try Config().tartTmpDir.appendingPathComponent("run-disk-\(UUID().uuidString)")
 
         try FileManager.default.copyItem(at: vmDir.diskURL, to: clonedDiskURL)
--- a/Sources/tart/Commands/Run.swift
+++ b/Sources/tart/Commands/Run.swift
@@ -67,12 +67,16 @@ struct Run: AsyncParsableCommand {
   var vncExperimental: Bool = false
 
   @Option(help: ArgumentHelp("""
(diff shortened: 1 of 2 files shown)


Sources/tart/Commands/Run.swift
```diff
@@ -495,6 +499,25 @@ struct Run: AsyncParsableCommand {
         continue
       }
 
+      // Support remote VM names in --disk command-line argument
+      if let remoteName = try? RemoteName(diskPath) {
+        let vmDir = try VMStorageOCI().open(remoteName)
+
+        // Unfortunately, VZDiskImageStorageDeviceAttachment does not support
+        // FileHandle, so we can't easily clone the disk, open it and unlink(2)
+        // to simplify the garbage collection, so use the <>
```
fkorotkov, line 508: Can we mount the same disk to two VMs? Maybe only in read-only mode?

edigaryev: Yes, but that will violate the OCI cache immutability guarantees that we currently have and still doesn't sound very secure compared to having a separate FS object.

Sources/tart/Commands/Run.swift
```diff
@@ -495,6 +499,25 @@ struct Run: AsyncParsableCommand {
...
       }
 
+      // Support remote VM names in --disk command-line argument
+      if let remoteName = try? RemoteName(diskPath) {
+        let vmDir = try VMStorageOCI().open(remoteName)
+
+        // Unfortunately, VZDiskImageStorageDeviceAttachment does not support
+        // FileHandle, so we can't easily clone the disk, open it and unlink(2)
+        // to simplify the garbage collection, so use an intermediate directory.
+        let clonedDiskURL = try Config().tartTmpDir.appendingPathComponent("run-disk-\(UUID().uuidString)")
```
fkorotkov, line 509: Should we skip it in `ro` mode?

edigaryev: Answered in https://github.com/cirruslabs/tart/pull/847#discussion_r1653093093.

fkorotkov (approved):

PR #864, opened 2024-07-15
tart pull: try to re-use APFS blocks by cloning the base image

edigaryev:
To be honest, I'm not a fan of this approach, because it's incredibly hard to measure the real-world impact of this, both in terms of CPU overhead and disk usage savings<sup>1</sup>.

The reason being that these variables are highly influenced by the contents of the base VM image (from which we're de-duplicating) and by the contents of the VM image to be pulled, yet we only have a few publicly available to run the tests on.

<sup>1</sup>: It feels that in some cases even losses, because we can't skip writing a zero block if the file we're basing off contains some other data in that place.

### DIFF
--- a/Sources/tart/OCI/Layerizer/DiskV2.swift
+++ b/Sources/tart/OCI/Layerizer/DiskV2.swift
@@ -1,5 +1,6 @@
 import Foundation
 import Compression
+import System
 
 class DiskV2: Disk {
   private static let bufferSizeBytes = 4 * 1024 * 1024
@@ -175,7 +176,13 @@ class DiskV2: Disk {
         try rdisk.seek(toOffset: offset)
         let actualContentsOnDisk = try rdisk.read(upToCount: chunk.count)
 
-        if chunk != actualContentsOnDisk {
+        if chunk == zeroChunk {
+          var arg = fpunchhole_t(fp_flags: 0, reserved: 0, fp_offset: off_t(offset), fp_length: off_t(chunk.count))
+
+          if fcntl(disk.fileDescriptor, F_PUNCHHOLE, &arg) == -1 {
+            throw RuntimeError.PullFailed("failed to punch hole: \(Errno.lastErrnoValue)")
+          }
+        } else if chunk != actualContentsOnDisk {
           try disk.seek(toOffset: offset)
           disk.write(chunk)
         }
--- a/Sources/tart/OCI/Layerizer/DiskV2.swift
+++ b/Sources/tart/OCI/Layerizer/DiskV2.swift
@@ -68,6 +68,15 @@ class DiskV2: Disk {
     try disk.truncate(atOffset: uncompressedDiskSize)
     try disk.close()
 
+    // Determine the file system block size
+    var st = stat()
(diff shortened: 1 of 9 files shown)


Sources/tart/OCI/Layerizer/DiskV2.swift
```diff
@@ -152,7 +169,25 @@ class DiskV2: Disk {
...
-      // Only write chunks that are not zero
+      // If the local layer cache is used, only write chunks that differ
+      // since the base disk can contain anything at any position
+      if let rdisk = rdisk {
+        try rdisk.seek(toOffset: offset)
+        let actualContentsOnDisk = try rdisk.read(upToCount: chunk.count)
+
+        if chunk != actualContentsOnDisk {
+          try disk.seek(toOffset: offset)
+          disk.write(chunk)
```
fkorotkov, line 180: It seems we can validate that the new chunk are zeroes here and use `g` syscall to erase a range. Here is an example that might work:

https://github.com/niw/HolePunch/blob/0f6716324915ddf073acb37e1dc268f38b071d3e/Sources/HolePunch/Command.swift#L76-L84

fkorotkov (commented):
IMO seems like a great improvement especially since there is a way to punch a hole in a file. I just pulled two images using the latest released version of Tart and one from this branch. I pulled `ghcr.io/cirruslabs/macos-sonoma-vanilla:latest` and `ghcr.io/cirruslabs/macos-sonoma-base:latest` which is basically vanilla with couple brew installes.

With the released version of Tart these two images took 42.03 GB of space and with this change (without punch hole thing) it takes 41.63 GB which seems to liitle of a win.

PS will be great to log that Tart found a base image to verify the logic and give a user some feedback.

Sources/tart/OCI/Layerizer/DiskV2.swift
edigaryev: Implemented in https://github.com/cirruslabs/tart/pull/864/commits/c110158b6ce7e93c21c35d16dbd564543f5eba8b.

edigaryev:
> PS will be great to log that Tart found a base image to verify the logic and give a user some feedback.

Implemented in https://github.com/cirruslabs/tart/pull/864/commits/a346ea67035e4926eb4478d8f5cb9311d618b536.

Sources/tart/VMStorageOCI.swift
```diff
@@ -200,8 +200,18 @@ class VMStorageOCI: PrunableStorage {
           // Choose the best base image which has the most deduplication ratio
           let localLayerCache = try await chooseLocalLayerCache(name, manifest, registry)
 
+          if let llc = localLayerCache {
```
fkorotkov, line 203: Let's add a check that `llc.deduplicatedBytes` is more than 0.

edigaryev: Added a check in https://github.com/cirruslabs/tart/pull/864/commits/a053c6b981202a703beb6c5c1804b94d28c1534b.

fkorotkov (approved):

PR #870, opened 2024-07-22
APFS reuse UX/DX improvements

fkorotkov:
Improvement to the APFS deduplication logic which checks whether a disk image file `mayShareFileContent` with some other file, and then we put a custom attribute to track the deduplication since there is no way to get this information from APFS itself.

It's not 100% accurate but given that OCI cache is immutable the actual disk usage can only be lover than that.

### DIFF
--- a/Sources/tart/URL+Prunable.swift
+++ b/Sources/tart/URL+Prunable.swift
@@ -25,16 +25,18 @@ extension URL: Prunable {
   }
 
   func setDeduplicatedBytes(_ size: UInt64) {
-    var intVal = size
-    let data = Data(bytes: &intVal, count: MemoryLayout.size(ofValue: intVal))
-    try! self.setExtendedAttribute(name: "user.deduplicatedBytes", value: data)
+    let data = "\(size)".data(using: .utf8)!
+    try! self.setExtendedAttribute(name: "run.tart.deduplicated-bytes", value: data)
   }
 
   func deduplicatedBytes() -> UInt64 {
-    let data = try? self.extendedAttributeValue(forName: "user.deduplicatedBytes")
-    if data == nil {
+    let data = try? self.extendedAttributeValue(forName: "run.tart.deduplicated-bytes")
+    guard let data else {
       return 0
     }
-    return data!.withUnsafeBytes { $0.load(as: UInt64.self) }
+    if let strValue = String(data: data, encoding: .utf8) {
+      return UInt64(strValue) ?? 0
+    }
+    return 0
   }
 }
--- a/Sources/tart/URL+Prunable.swift
+++ b/Sources/tart/URL+Prunable.swift
@@ -30,8 +30,7 @@ extension URL: Prunable {
   }
 
   func deduplicatedBytes() -> UInt64 {
-    let data = try? self.extendedAttributeValue(forName: "run.tart.deduplicated-bytes")
-    guard let data else {
+    guard let data = try? self.extendedAttributeValue(forName: "run.tart.deduplicated-bytes") else {
       return 0
     }
     if let strValue = String(data: data, encoding: .utf8) {
--- a/Sources/tart/URL+Prunable.swift
+++ b/Sources/tart/URL+Prunable.swift
@@ -11,13 +11,17 @@ extension URL: Prunable {
   }
 
   func allocatedSizeBytes() throws -> Int {
+    try resourceValues(forKeys: [.totalFileAllocatedSizeKey]).totalFileAllocatedSize!
+  }
+
+  func deduplicatedSizeBytes() throws -> Int {
     let values = try resourceValues(forKeys: [.totalFileAllocatedSizeKey, .mayShareFileContentKey])
     // make sure the file's origin file is there and duplication works
     var dedublicatedSize = 0
     if values.mayShareFileContent == true {
-      dedublicatedSize = Int(deduplicatedBytes())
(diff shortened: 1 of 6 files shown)


fkorotkov:
With that change `tart list` will look like this:

```
Source Name                                                                                                            Disk Size State  
OCI    ghcr.io/cirruslabs/macos-sonoma-base:latest                                                                     50   9    stopped
OCI    ghcr.io/cirruslabs/macos-sonoma-base@sha256:2ddafad3147c46330d6b13112b65113b529eb0c7ab6d4692168e8417c4678862    50   9    stopped
OCI    ghcr.io/cirruslabs/macos-sonoma-vanilla:latest                                                                  50   18   stopped
OCI    ghcr.io/cirruslabs/macos-sonoma-vanilla@sha256:b4b6adc2598cd479e1cc9d58c289d8931749bd80a67e8332cce2216d0d813ca7 50   18   stopped
```

Instead of

```
Source Name                                                                                                            Disk Size State  
OCI    ghcr.io/cirruslabs/macos-sonoma-base:latest                                                                     50   23   stopped
OCI    ghcr.io/cirruslabs/macos-sonoma-base@sha256:2ddafad3147c46330d6b13112b65113b529eb0c7ab6d4692168e8417c4678862    50   23   stopped
OCI    ghcr.io/cirruslabs/macos-sonoma-vanilla:latest                                                                  50   18   stopped
OCI    ghcr.io/cirruslabs/macos-sonoma-vanilla@sha256:b4b6adc2598cd479e1cc9d58c289d8931749bd80a67e8332cce2216d0d813ca7 50   18   stopped
```

edigaryev:
I'm not sure if hiding the actual allocated size that Finder, `ls`, `du` and other tools report is a good idea.

What do you think about adding a new column instead?

Like this:

```
Source Name                                                                                                         Disk Size Size (CoW) State  
OCI    127.0.0.1:8080/macos:latest                                                                                  50   23   11         stopped
OCI    127.0.0.1:8080/macos@sha256:30de8331acc5c1007f4b84b10abde8217e820c5842f237963c461f94037bab7e                 50   23   23         stopped
OCI    ghcr.io/cirruslabs/macos-sonoma-base@sha256:0222d6863142d10095fdd64845a19781347a1f3d232a771c6e2ec143ca9c1ac5 50   24   24         stopped
```

Sources/tart/URL+Prunable.swift
```diff
@@ -10,10 +11,30 @@ extension URL: Prunable {
...
   }
 
   func sizeBytes() throws -> Int {
     try resourceValues(forKeys: [.totalFileSizeKey]).totalFileSize!
   }
+
+  func setDeduplicatedBytes(_ size: UInt64) {
+    var intVal = size
+    let data = Data(bytes: &intVal, count: MemoryLayout.size(ofValue: intVal))
+    try! self.setExtendedAttribute(name: "user.deduplicatedBytes", value: data)
```
edigaryev, line 30: Should we use something like `org.cirruslabs.tart.deduplicated-bytes` instead to avoid confusion in the future, and make this variable a constant?

Sources/tart/URL+Prunable.swift
```diff
@@ -10,10 +11,30 @@ extension URL: Prunable {
...
+    return values.totalFileAllocatedSize! - dedublicatedSize
   }
 
   func sizeBytes() throws -> Int {
     try resourceValues(forKeys: [.totalFileSizeKey]).totalFileSize!
   }
+
+  func setDeduplicatedBytes(_ size: UInt64) {
+    var intVal = size
+    let data = Data(bytes: &intVal, count: MemoryLayout.size(ofValue: intVal))
```
edigaryev, line 29: What do you think about storing the values as strings instead?

This has two benefits:

* no need to bother with endianness
* `xattr -l` will be able to show the value that we set, which aids in debugging

Sources/tart/URL+Prunable.swift
```diff
@@ -10,10 +11,30 @@ extension URL: Prunable {
...
+    try! self.setExtendedAttribute(name: "user.deduplicatedBytes", value: data)
+  }
+
+  func deduplicatedBytes() -> UInt64 {
+    let data = try? self.extendedAttributeValue(forName: "user.deduplicatedBytes")
+    if data == nil {
+      return 0
+    }
```
edigaryev, line 37: What do you think about simplifying this with a `guard` statement?

Sources/tart/URL+Prunable.swift
fkorotkov: I read somewhere that custom attributes should start with `user.` but seems we can do something like `run.tart.deduplicated-bytes`

edigaryev: Not sure about `run.tart.*` as we already use `org.cirruslabs.tart.*` here: https://github.com/cirruslabs/tart/blob/4ed73bc7753a491b22208a13f49b4517cc6af97d/Sources/tart/OCI/Manifest.swift#L14-L15

Sources/tart/URL+Prunable.swift
fkorotkov: SGTM. See [a42174c](https://github.com/cirruslabs/tart/pull/870/commits/a42174c780ec8a468ae805700c4b66ad3bc0a47c)

edigaryev (approved):

PR #875, opened 2024-07-31
Support setting root disk synchronization mode

nfd:
Following on from https://github.com/cirruslabs/tart/pull/872, also support setting VZDiskImageSynchronizationMode for /.

This is useful because it significantly improves write performance in situations where the increased potential for data loss doesn't matter.

### DIFF
--- a/Sources/tart/Commands/Get.swift
+++ b/Sources/tart/Commands/Get.swift
@@ -1,6 +1,5 @@
 import ArgumentParser
 import Foundation
-import Virtualization
 
 fileprivate struct VMInfo: Encodable {
   let OS: OS
--- a/Sources/tart/Commands/Get.swift
+++ b/Sources/tart/Commands/Get.swift
@@ -7,7 +7,6 @@ fileprivate struct VMInfo: Encodable {
   let Memory: UInt64
   let Disk: Int
   let Size: String
-  let Sync: String
   let Display: String
   let Running: Bool
   let State: String
@@ -27,7 +26,7 @@ struct Get: AsyncParsableCommand {
     let vmConfig = try VMConfig(fromURL: vmDir.configURL)
     let memorySizeInMb = vmConfig.memorySize / 1024 / 1024
 
-    let info = VMInfo(OS: vmConfig.os, CPU: vmConfig.cpuCount, Memory: memorySizeInMb, Disk: try vmDir.sizeGB(), Size: String(format: "%.3f", Float(try vmDir.allocatedSizeBytes()) / 1000 / 1000 / 1000), Sync: vmConfig.sync.description, Display: vmConfig.display.description, Running: try vmDir.running(), State: try vmDir.state().rawValue)
+    let info = VMInfo(OS: vmConfig.os, CPU: vmConfig.cpuCount, Memory: memorySizeInMb, Disk: try vmDir.sizeGB(), Size: String(format: "%.3f", Float(try vmDir.allocatedSizeBytes()) / 1000 / 1000 / 1000), Display: vmConfig.display.description, Running: try vmDir.running(), State: try vmDir.state().rawValue)
     print(format.renderSingle(info))
   }
 }
--- a/Sources/tart/Commands/Get.swift
+++ b/Sources/tart/Commands/Get.swift
@@ -1,5 +1,6 @@
 import ArgumentParser
 import Foundation
+import Virtualization
 
 fileprivate struct VMInfo: Encodable {
   let OS: OS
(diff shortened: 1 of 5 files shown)


edigaryev:
What do you think about a `--sync` argument to `tart run` instead that takes `none`, `fsync` and `full` and is `full` by default?

I'm not sure if the root disk synchronization mode is something that we should persist when pushing to the VM image to the OCI registry.

nfd:
That makes sense...

fkorotkov (approved):
Thank you for finding this out? Have you measured the performance gains?

We definitely need to check setting it for macOS ephemeral VMs which are also fine with corruption on exit. :-)

nfd:
I haven't tested it extensively -- but for my workload, which is sending SQL to mariadb in a docker container, it's about a 17x speedup:

fsync: 786ms, 781ms, 528ms, avg 698 ms

none: 11.89s, 11.72s, 12.47s, avg 12 seconds

I don't see why it wouldn't work for MacOS as well, I just noticed that MacOS clients weren't being sent a VZDiskImageSynchronizationMode at all whereas Linux clients were.

nfd:
By the way, I don't think this makes a difference for macOS images.

I edited the `VZDiskImageStorageDeviceAttachment` set-up in VM.swift to use the command-line-supplied caching mode for macOS as well, and I measured the same time (about 4 seconds) for this benchmark for both --sync=full and --sync=none:

```
import os

with open('testfile.dat', 'wb') as h:
    for i in range(100):
        for j in range(1024):
            h.write(b'*' * 4096)
            h.flush()
            os.fsync(h.fileno())
```

On Linux debian, this takes about 18 seconds with --sync=none, and at least two minutes with --sync=full (I got bored and killed it).

`VZDiskImageStorageDeviceAttachment` also accepts a `cachingMode` and this didn't make a difference on Sonoma either.

So I think macOS guests either behave differently or are treated differently such that you can't enable full synchronization, in which case this PR will only benefit Linux guests.

edigaryev:
> By the way, I don't think this makes a difference for macOS images.

Another thing to note is that there seems to be little to none performance improvement for all the other non-DB software that does not rely on `sync(2)`/`fsync(2)`, even on Linux. [Checked this using `fio`](https://github.com/cirruslabs/tart/blob/main/benchmark/internal/command/fio/benchmark.go) with and without the `--sync=1` argument on `ghcr.io/cirruslabs/ubuntu:latest`.

In your case the speed-up can also be accomplished by e.g. [reducing MySQL durability for testing](http://www.tocker.ca/2013/11/04/reducing-mysql-durability-for-testing.html) to avoid introducing a new command-line argument in `tart run`, which is already overloaded, as you can see.

And in case we're going to introduce it, we probably need to think about how to support setting a different [caching mode](https://developer.apple.com/documentation/virtualization/vzdiskimagecachingmode) too, perhaps by naming this argument as something like `--root=options`, where `options` will have the similar format as the `--disk` argument options.

nfd:
If you think `tart run` is overloaded then why did you suggest I change the PR so that it added a flag to `tart run`?

I don't think it's appropriate to modify the configuration of containers I'm working with (which are built to be deployed to production) simply to work around an issue with virtualisation.

edigaryev:
> If you think `tart run` is overloaded then why did you suggest I change the PR so that it added a flag to `tart run`?

Because I've initially thought that the performance without `none` synchronization is universally bad and that there has to be a way to change this somehow, yet it's more desirable to provide such an way explicitly when invoking `tart run` to avoid surprises for people/robots who just `tart run` a VM that was accidentally pushed with behavior `tart set`'d.

But then it turned out to be bad only when `sync(2)`/`fsync(2)` is used, which means that this issue can alternatively be solved on the guest's side too. And also by attaching a secondary disk using `--disk` and setting an`sync=none` option.

Another realization I've got after modifying Tart to run the `fio` benchmark I've mentioned above is that we don't allow setting the `cachingMode` dynamically, thus it makes sense to least make the new command-line argument a bit more flexible to support both the `synchronizationMode` and `cachingMode`, even if the `synchronizationMode` support is not done as a part of this PR.

nfd:
OK, I understand the broader picture. I think it makes sense to expose this option from tart rather than forcing the slowest (and to be honest probably least desirable, for many workloads) choice upon the user. I'm not fussed about how you do it. Adding caching mode selection as well is beyond the scope of this particular PR, though if you've got an idea about how it would be added I'd be happy to add it in a second one.

By the way, I think fsync() being slow affects more than just obvious database loads. For example, sqlite fsync()s and is embedded by many programs. Pure filesystem benchmarks are useful but not the whole story.

edigaryev (approved):

PR #878, opened 2024-08-05
Optimize DiskV2 Deduplication

fkorotkov:
I ran some Instruments CPU profiles pulling `ghcr.io/cirruslabs/macos-sonoma-base:latest` on top of `ghcr.io/cirruslabs/macos-sonoma-vanilla:latest` and found a few bottle necks. See commits for details.


Sources/tart/OCI/Layerizer/DiskV2.swift
```diff
@@ -198,14 +209,22 @@ class DiskV2: Disk {
...
       // If the local layer cache is used, only write chunks that differ
       // since the base disk can contain anything at any position
       if let rdisk = rdisk {
-        // F_PUNCHHOLE requires the holes to be aligned to file system block boundaries
-        let isHoleAligned = (offset % fsBlockSize) == 0 && (UInt64(chunk.count) % fsBlockSize) == 0
+        try rdisk.seek(toOffset: offset)
+        let actualContentsOnDisk = try rdisk.read(upToCount: chunk.count)
 
-        if isHoleAligned && chunk == zeroChunk {
+        if actualContentsOnDisk == chunk {
```
fkorotkov, line 219: This is still a place for optimization. One though was to not try to optimize wholes on the sub-layer level, only try to deduplicate layers. And in this case whole punch the whole uncompressed layer and then sparse write into it.

Sources/tart/OCI/Layerizer/DiskV2.swift
```diff
@@ -121,11 +121,14 @@ class DiskV2: Disk {
         // Launch a fetching and decompression task
         group.addTask {
           // No need to fetch and decompress anything if we've already done so
-          if try pullResumed && Digest.hash(diskURL, offset: diskWritingOffset, size: uncompressedLayerSize) == uncompressedLayerContentDigest {
-            // Update the progress
-            progress.completedUnitCount += Int64(diskLayer.size)
+          if pullResumed {
```
edigaryev, line 124: I'm not sure if we need this change, because currently [short-circuit evaluation](https://en.wikipedia.org/wiki/Short-circuit_evaluation) will kick in if `pullResumed` is `false`, and hash won't be calculated.

fkorotkov: Not according to the profiling I did. Both parts of the expression are getting evaluated regardless.

Sources/tart/Data+Chunks.swift
```diff
@@ -0,0 +1,18 @@
+import Foundation
+
+extension Data {
+  /*
+   * Performant version of splitting a Data into chunks of a given size.
+   * It appers that "Data.chunks` is not as performant as chunking the range of the data
```
edigaryev, line 6: How did you measure this? Have you used "CPU Profiler" of Instruments? Have you ran `sudo purge` before the test?

I've replaced the `tart list`'s `run()` function with the following contents:

```swift
let disk = try FileHandle(forReadingFrom: URL(fileURLWithPath: "/Users/edi/.tart/vms/macos/disk.img"))

let holeGranularityBytes = 64 * 1024

var count = 0

while true {
  guard let data = try disk.read(upToCount: 64 * 1024 * 1024) else {
    break
  }
  
  for chunk in data.chunks(ofCount: holeGranularityBytes) {
    count += chunk.count
  }
}

print(count)
```

Here's a profile of `chunks()` which takes ~15 seconds and ~20 Mc to churn through macOS Sonoma's disk image:

Screenshot 2024-08-05 at 15 58 08

Now by changing `chunks()` to `subdataChunks()` we do the same with ~27 seconds and using ~110 Mc:

Screenshot 2024-08-05 at 16 00 17

Sources/tart/OCI/Layerizer/DiskV2.swift
edigaryev: Is it possible that `pullResumed` was set to `true` when the profiling was made?

fkorotkov: Unlikely, I was running pulls to completion between profiles.

Sources/tart/Data+Chunks.swift
fkorotkov: I didn't purge but this was not the biggest optimization. Let's revert and profile separately. See [773a403](https://github.com/cirruslabs/tart/pull/878/commits/773a40301136482a9f7103048f8441b7bb63b429)

Sources/tart/OCI/Layerizer/DiskV2.swift
edigaryev: I've just done some debugging and I think that we can safely revert this because the compiler is correct w.r.t. this piece of code.

It's just that there's one more call to the `Digest.hash()` that is visible in the Instruments "CPU Profiler", and it is located below:

https://github.com/cirruslabs/tart/blob/33b5cfe2ed4b74e35ab60eefc53aede25b54db18/Sources/tart/OCI/Layerizer/DiskV2.swift#L143

fkorotkov: On Sequoia it's still the case. Maybe something to do about the `try`, don't know.

edigaryev (approved):

PR #900, opened 2024-09-09
tart run: provide a hint with names of other running VMs

edigaryev:
When VM limit gets exceeded.


fkorotkov (approved):
I thought about the same! But on Cirrus CLI level. This is even better!

PR #904, opened 2024-09-16
Completely disable audio devices in case of `--no-audio`

fkorotkov:
With the current behaviour VMs do have audio devices but they don't pass-through signal to host which can lead to issues like https://github.com/actions/runner-images/issues/9330 . 

With this change there will be just no audio at all but commands like `say "hello"` work fine. So people don't need to apply the workaround for the permission dialog at all.

### DIFF
--- a/Sources/tart/Commands/Run.swift
+++ b/Sources/tart/Commands/Run.swift
@@ -70,7 +70,7 @@ struct Run: AsyncParsableCommand {
   @Flag(help: ArgumentHelp("Force open a UI window, even when VNC is enabled.", visibility: .private))
   var graphics: Bool = false
 
-  @Flag(help: "Disable audio pass-through to host.")
+  @Flag(help: "Do not configure audio device.")
   var noAudio: Bool = false
 
   @Flag(help: ArgumentHelp(
--- a/Sources/tart/Commands/Run.swift
+++ b/Sources/tart/Commands/Run.swift
@@ -70,7 +70,7 @@ struct Run: AsyncParsableCommand {
   @Flag(help: ArgumentHelp("Force open a UI window, even when VNC is enabled.", visibility: .private))
   var graphics: Bool = false
 
-  @Flag(help: "Do not configure audio device.")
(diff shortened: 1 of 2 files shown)


edigaryev:
It seems that this permission issue is only related to the audio input (microphone, for security/privacy reasons).

This means that we can keep the audio output intact when `--no-audio` is enabled, thus still covering the use-case of providing a dummy audio output device for some people:

```patch
diff --git a/Sources/tart/VM.swift b/Sources/tart/VM.swift
index ba3e8fa..8dbdc88 100644
--- a/Sources/tart/VM.swift
+++ b/Sources/tart/VM.swift
@@ -316,16 +316,18 @@ class VM: NSObject, VZVirtualMachineDelegate, ObservableObject {
 
     // Audio
     let soundDeviceConfiguration = VZVirtioSoundDeviceConfiguration()
-
-    let inputAudioStreamConfiguration = VZVirtioSoundDeviceInputStreamConfiguration()
     let outputAudioStreamConfiguration = VZVirtioSoundDeviceOutputStreamConfiguration()
 
     if audio && !suspendable {
+      let inputAudioStreamConfiguration = VZVirtioSoundDeviceInputStreamConfiguration()
       inputAudioStreamConfiguration.source = VZHostAudioInputStreamSource()
       outputAudioStreamConfiguration.sink = VZHostAudioOutputStreamSink()
+
+      soundDeviceConfiguration.streams = [inputAudioStreamConfiguration, outputAudioStreamConfiguration]
+    } else {
+      soundDeviceConfiguration.streams = [outputAudioStreamConfiguration]
     }
 
-    soundDeviceConfiguration.streams = [inputAudioStreamConfiguration, outputAudioStreamConfiguration]
     configuration.audioDevices = [soundDeviceConfiguration]
 
     // Keyboard and mouse
```

edigaryev (approved):

fkorotkov:
OK. The description of `--no-audio` is a a bit off IMO but I guess this way it's a less of a breaking change.

PR #915, opened 2024-10-01
Make `ARPCache` to respect cancellation

fkorotkov:
By manually handling `waitUntilExit`

Tested by using `/bin/sleep 120` instead of `/usr/sbin/arp -an`.

### DIFF
--- a/Sources/tart/MACAddressResolver/ARPCache.swift
+++ b/Sources/tart/MACAddressResolver/ARPCache.swift
@@ -41,7 +41,7 @@ struct ARPCacheInternalError: Error, CustomStringConvertible {
 struct ARPCache {
   let arpCommandOutput: Data
 
-  init() throws {
+  init() async throws {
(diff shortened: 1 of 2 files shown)


Sources/tart/MACAddressResolver/ARPCache.swift
```diff
@@ -52,7 +52,12 @@ struct ARPCache {
     process.standardInput = FileHandle.nullDevice
 
     try process.run()
-    process.waitUntilExit()
+    while process.isRunning {
+      if Task.isCancelled {
+        process.terminate()
```
edigaryev, line 57: It seems to me that this line won't be called 99.99% of the time, because `Task.sleep()` runs long enough and will throw a `CancellationError`, thus exiting this loop.

What do you think about addressing the root cause instead? See https://github.com/cirruslabs/tart/pull/916.

PR #924, opened 2024-10-25
tart {clone,pull}: make deduplication opt-in

edigaryev:
As noted in https://github.com/cirruslabs/tart/pull/864, deduplication is a somewhat controversial feature that might result in noticeably slower pulls with larger network speeds (think 10 Gbps) due to disk reads and comparisons needed.

Also, hole punching mechanism used for deduplication can presumably result in APFS performance degradation later, when running the cloned VMs.

On the other hand, typical disk space savings observed for Cirrus Labs images are around 12 GB per pulled VM, which might not worth the extra complexity involved and potential performance implications.

### DIFF
--- a/Sources/tart/Commands/Clone.swift
+++ b/Sources/tart/Commands/Clone.swift
@@ -28,6 +28,9 @@ struct Clone: AsyncParsableCommand {
   @Option(help: "network concurrency to use when pulling a remote VM from the OCI-compatible registry")
   var concurrency: UInt = 4
 
+  @Flag(help: .hidden)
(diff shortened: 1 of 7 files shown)


edigaryev:
> Should we just don't call `chooseLocalLayerCache` at all if the deduplication flag is false instead of passing `deduplicate` everywhere?

In this case we won't re-use the chunks from other VM images in the OCI cache if there's a match.

This behavior was introduced by https://github.com/cirruslabs/tart/pull/825.

Why give it up?

PR #925, opened 2024-10-25
Update access time on `pull`

fkorotkov:
To make sure we won't prune then immediately after. Useful for when scenarios similar to Cirrus CLI when we make sure that several images are up-to-date before every request for task execution.


edigaryev:
Indeed, good catch!

edigaryev (approved):

PR #931, opened 2024-10-31
utimes(2): use errno to explain the error

edigaryev:
Otherwise we get some nonsense from `SecCopyErrorMessageString()` in `explanation()` when the error occurs.

Related to `PERSISTENT-WORKERS-43`.

### DIFF
--- a/Sources/tart/URL+AccessDate.swift
+++ b/Sources/tart/URL+AccessDate.swift
@@ -1,4 +1,5 @@
 import Foundation
+import System
 
 extension URL {
   func accessDate() throws -> Date {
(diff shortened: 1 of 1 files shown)


hblockx:
Soooo the utimes(2) failed: Operation not permitted - error is known? I cannot clone with tart anymore neither does it work with orchard...

edigaryev:
> Soooo the utimes(2) failed: Operation not permitted - error is known? I cannot clone with tart anymore neither does it work with orchard...

1. Which command-line do you use to invoke the Tart?

2. Is it reproducible?

3. Do you use `TART_HOME`?

4. Have you ran Tart under `root`? It could've messed your `~/.tart` permissions.

hblockx:
yeah that is exactly what I noticed right now. I was using orchard with tart via daemon as a normal user. This broke with macOS 15.1, maybe because of the missing local network permission. Therefore we switched temporary to root access and this messed up .tart, as we were setting TART_HOME so the root user does not pull the images again and again... I am trying to fix .tart with chmod atm.

hblockx:
Seems a bit like tart has no permission to create the bridge to 192.168.64.XXX anymore. On older macOS-Versions the bridge is there and it is still working, on new Macs it is not setting it up

Issue #928, opened 2024-10-30
UI window no longer opens after updating to Sequoia 15.1


phillipross:
I'm not sure if I'm missing something or bumping into a bug, but I no longer get a UI window opening when I `tart run` a macOS VM or Linux VM.  The VMs start and I can ssh to them, but simply no UI window appears.

phillipross:
I did find an icon in the dock, and clicking on it did open up the UI window, thanks!

Is this a functionality change in tart or macOS or a combination?  Thanks again!

phillipross:
If nothing in tart has changed then I imagine it's more of a Sequoia thing.  This behavior is very persistent, behaving this way on 3 different Sequoia installs I have now, 1 having been upgraded from Sonoma in-place and the others freshly installed from scratch w/o using migration assistant.  Unfortunately I didn't get to test on Sequoia 15.0.1 before updating to 15.1 so can't actually isolate it to that granularity, but this new behavior is consistent for me.  I'm happy to provide additional details/tests if it's helpful. Thanks again!

edigaryev:
How are you installing Tart, by the way?

I'm observing the same behavior when running Tart from the downloaded `.tar.gz`, but not from `brew`.

vitorgalvao:
I am seeing this issue consistently on macOS 15.3.2 and tart 2.24.1, installed from the official formula.

It has been happening for a while. At first I thought I might just have had other windows in the way, but I’ve made sure that is not the case.

Starting a VM with `tart run` works, but I do not see the window until I click the icon in the dock. Searching more on the repo indicates this should’ve been fixed in [2.20.0](https://github.com/cirruslabs/tart/releases/tag/2.20.0), published by @fkorotkov, after the merge of https://github.com/cirruslabs/tart/pull/935 authored by @edigaryev, but I’m still seeing it several versions later.

I can open this as a new issue if you prefer, but this one seemed appropriate.

edigaryev:
@vitorgalvao 

1. How do you invoke the `tart run` (e.g. from a Terminal app)
2. Which arguments do you pass to `tart run`?

Also, could you please try with latest macOS 15.4? It was released just a day ago.

vitorgalvao:
[...]

PR #935, opened 2024-11-04
Info.plist: do not use LSBackgroundOnly

edigaryev:
This was introduced in https://github.com/cirruslabs/tart/pull/746, but I've tested this change with `tart create --from-ipsw=latest <VM>`, and no icon in Dock was shown.

Resolves https://github.com/cirruslabs/tart/issues/928.


torarnv:
Thanks! 🙏 This stuff is unfortunately very shaky on the OS level, as https://stackoverflow.com/a/12981068/9316533 shows 😢

PR #938, opened 2024-11-06
`tart run --nested` to enable nested virtualization when available

fkorotkov:
Only works for Linux VMs under Sequoia hosts.

Fixes #933
Fixes #701

### DIFF
--- a/Sources/tart/Platform/Darwin.swift
+++ b/Sources/tart/Platform/Darwin.swift
@@ -6,12 +6,6 @@ struct UnsupportedHostOSError: Error, CustomStringConvertible {
   }
 }
 
-struct UnsupportedNestedVirtualization: Error, CustomStringConvertible {
-  var description: String {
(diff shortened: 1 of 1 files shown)


Sources/tart/Platform/Darwin.swift
```diff
@@ -58,7 +64,11 @@ struct UnsupportedHostOSError: Error, CustomStringConvertible {
       VZMacOSBootLoader()
     }
 
-    func platform(nvramURL: URL) throws -> VZPlatformConfiguration {
+    func platform(nvramURL: URL, needsNestedVirtualization: Bool) throws -> VZPlatformConfiguration {
+      if needsNestedVirtualization {
+        throw UnsupportedNestedVirtualization()
```
edigaryev, line 69: What do you think about throwing `RuntimeError.VMConfigurationError` here with an appropriate message, instead of introducing a new `Error` struct?

edigaryev (approved):

Issue #682, opened 2023-12-05
GRUB too high resolution


unikzforce:
Hello Tart team,

can you please guide me to understand why the resolution is too high on the grub menu? it will be fixed afterwards.

is this a bug?

![image](https://github.com/cirruslabs/tart/assets/7010566/538ebc9a-0dfd-4e98-9833-f13257654397)

after booting:
![image](https://github.com/cirruslabs/tart/assets/7010566/cd6dfcea-aaa1-403a-a3d2-ca8e7ecf71ce)

fkorotkov:
Could you please try with the official Tart installation?

'''
brew uninstall tart
brew install cirruslabs/cli/tart
'''

unikzforce:
sure, i have tried unfortunately it would stay the same same. 
(by the way i'm using Sonoma 14.1.1)

fkorotkov:
If I understand you issue correctly, display size for your VM is too big and therefore everything is too small in the window?

Can you try running `tart get fedora`? It will print out display size and you can change it via `tart set --display 1200x800 fedora`.

unikzforce:
i have tested many resolutions, even 200x80, there is no change in the result

![image](https://github.com/cirruslabs/tart/assets/7010566/2ccc8668-51a9-4446-b2db-93dc071aad79)

unikzforce:
in my sole laptop without any external monitor, both the grub and the booted fedora have very small fonts.

if i use my external 34inch monitor, it's getting a bit more usable when the system is booted, but still things are barely visible in grub.

this pictures are taken when using external 34 inch monitor:

![image](https://github.com/cirruslabs/tart/assets/7010566/634864f4-5c85-416f-bd8d-caf90b5f7682)

unikzforce:
I think the scaling depends on the monitor that tart window is open on,
now the external monitor is opened, i transition the program to macbook 16inch lcd

![image](https://github.com/cirruslabs/tart/assets/7010566/619f191f-8617-4e08-bb6a-bba75b1c06a4)

and then without restarting the program, I move the tart window to the 34inch external monitor:
![image](https://github.com/cirruslabs/tart/assets/7010566/b8b6ae3b-9004-4141-ac04-3f97222038e4)

bossprocedure:
[...]

PR #951, opened 2024-11-18
Add flag to allow automatically reconfigure display for linux VM

Fred78290:
By default when using tart to create linux VM with graphical desktop, the resizing of the window don't resize linux desktop and doesn't permit fullscreen.

This pull request add a flag --allows-reconfigure-display to run command permitting machineView.automaticallyReconfiguresDisplay = true

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,6 @@ dist/
 
 # mkdocs-material
 site
+
+DerivedData
+.vscode
--- a/Sources/tart/Commands/Run.swift
+++ b/Sources/tart/Commands/Run.swift
@@ -201,6 +201,9 @@ struct Run: AsyncParsableCommand {
   #endif
   var captureSystemKeys: Bool = false
 
+  @Flag(name: [.customLong("allows-reconfigure-display")], help: ArgumentHelp("Allow to reconfigure display when window is resized for linux VM"))
+  var automaticallyReconfiguresDisplay: Bool = false
+
   mutating func validate() throws {
     if vnc && vncExperimental {
       throw ValidationError("--vnc and --vnc-experimental are mutually exclusive")
@@ -478,7 +481,7 @@ struct Run: AsyncParsableCommand {
 
       NSApplication.shared.run()
     } else {
-      runUI(suspendable, captureSystemKeys)
+      runUI(suspendable, captureSystemKeys, vm?.config.os == .darwin || automaticallyReconfiguresDisplay)
     }
   }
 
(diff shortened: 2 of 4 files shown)


edigaryev:
Hello Frederic 👋 

What do you think about having a `--display-auto-reconfigure` option in `tart set` instead?

`automaticallyReconfiguresDisplay` will be then set to a value that depends on the conditions below:

| Platform | displayAutoReconfigure in `config.json` | `automaticallyReconfiguresDisplay` |
|----------|-----------------------------------------|------------------------------------|
| macOS    | `null`<sup>1</sup>                      | `true`                             |
| macOS    | `true`                                  | `true`                             |
| macOS    | `false`                                 | `false`                            |
| Linux    | `null`<sup>1</sup>                      | `false`                            |
| Linux    | `true`                                  | `true`                             |
| Linux    | `false`                                 | `false`                            |

<sup>1</sup>: backwards compatibility with VM images created before `--display-auto-reconfigure` was implemented

I also propose to not populate the `displayAutoReconfigure` in `config.json` for now.

Fred78290:
> > The argument remains discrete for the user as nested flag :)
> 
> We cannot expose every possible option in `tart run`. Our strategy is try to avoid adding any additional options if the same can be achieved by a sane default.
> 
> Unfortunately, in this case, due to #682 it's not possible, so here we go.
> 
> > Else every options from run command must be moved to Config.json, such as network options or vnc options.
> 
> It's not black and white 😉
> 
> We try to keep security-sensitive options in `tart run`, so that a freshly cloned VM is secure and doesn't e.g. mount your home directory automatically on a first run.
> 
> But from the security point of view, `--display-auto-reconfigure` seems to be a relatively safe option.
> 
> Another point is that we already have `--display` option in `tart set`. Why not co-locate them, to have a lower cognitive load for the user?

I don't understand your logic, in my previous pull request I putted the nested option in config.json and you moved it as flag on run command, so now the kind of flag that i request looks like the same behavior and now you say to put it in config.json.

You gimme some trouble :)

edigaryev:
> I don't understand your logic, in my previous pull request I putted the nested option in config.json and you moved it as flag on run command, so now the kind of flag that i request looks like the same behavior and now you say to put it in config.json.

Please take a look at https://github.com/cirruslabs/tart/pull/951#issuecomment-2483821971, I've explained some of the reasoning there.

`--nested`, for example, has a direct impact on security (just Google `nested virtualization CVE`), so it makes sense to not enable it by default nor preserve it in the `config.json`, which will be propagated through the OCI registry.

Fred78290:
> `--nested`, for example, has a direct impact on security (just Google `nested virtualization CVE`), so it makes sense to not enable it by default nor preserve it in the `config.json`, which will be propagated through the OCI registry.

It's concern AMD with KVM not apple sicilon with VZ....

PR #954, opened 2024-11-19
tart set: support --{,no-}display-auto-reconfigure

edigaryev:
A simpler approach with direct `VMConfig.displayAutoReconfigure` access, inverted flag (to support prepending `--no-*`) and backwards compatibility (according to [the truth table](https://github.com/cirruslabs/tart/pull/951#issuecomment-2483611581)).

/cc @Fred78290

Resolves https://github.com/cirruslabs/tart/pull/951.

### DIFF
--- a/Sources/tart/Commands/Run.swift
+++ b/Sources/tart/Commands/Run.swift
@@ -740,12 +740,12 @@ struct VMView: NSViewRepresentable {
 
     machineView.capturesSystemKeys = capturesSystemKeys
 
-    // Enable automatic display reconfiguration
-    // for guests that support it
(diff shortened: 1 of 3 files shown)


edigaryev:
> I suggest --refit-display

Would `--display-refit` work for you?

To semantically group it with the `--display` option.

Fred78290:
@edigaryev 

It should be more simplier to use
```
@Flag(name: [.customLong("display-refit")]
```

edigaryev:
> It should be more simplier to use

Why do we need `.customLong` if the argument's name is already automatically derived from the property’s name?

PR #965, opened 2024-12-04
New benchmark results on AWS mac2.metal for Sonoma and Sequoia guests

edigaryev:
Also changed the installation instructions slightly to require passwordless sudo and run `sudo purge && sync` before each test.

### DIFF
--- a/benchmark/README.md
+++ b/benchmark/README.md
@@ -2,10 +2,18 @@
 
 Tart comes with a Golang-based benchmarking utility that allows one to easily compare host and guest performance.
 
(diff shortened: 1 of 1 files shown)


edigaryev:
> What's the verdict?

Using `sync=none` with `caching=cached` almost always improves read/write throughput on EBS volumes, except for sync-heavy workloads (but that could be a statistically insignificant data also).

However, there's no reliable way to detect that we're running on AWS: no specific environment variables and the AWS Instance Metadata Service is not always accessible.

Enabling `--root-disk-opts="sync=none,caching=cached"` when `CI` environment variable is set seems like a magic to me. I'd rather have this option separately documented in FAQ.

What do you think?

Issue #993, opened 2025-01-12
How to get Host MAC IP from inside virtual machine


jaykam:
Hi @fkorotkov 

I'm running virtual machine on Sonoma MAC Host and I'm able to get virtual IP from the Host using 

`tart ip <virtual-name> --resolver arp`

But is there any way to get Base Host IP from inside the virtual machine? I'm using bridging networking

edigaryev:
Hi @jaykam,

I don't think there's a readily available way to do that unless you somehow communicate either the host's local network address or a `.local` domain name to the VM.

The whole point of `--net-bridged` is that it largely detaches the networking decisions from the host and moves the responsibility to your local network's DHCP server.

You could make your host's address static on your local network's DHCP server, and then simply hardcode it in your VM provisioning scripts, for example, or try to resolve the host's IP address [through mDNS](https://support.apple.com/en-us/101903).

Note that when using the shared (NAT) networking, figuring out the host's IP address is much easier, as it's almost always the first address in a network assigned to the VM.

E.g. the host will have address `192.168.64.1` in network `192.168.64.0/24`.

Yesserre:
Closed #993 <https://github.com/cirruslabs/tart/issues/993> as completed.
Op di 14 jan 2025 om 21:35 schreef Nikolay Edigaryev <
***@***.***>

Issue #873, opened 2024-07-26
Can't resolve guest IP address


nre-ableton:
I'm having difficulty resolving IP addresses for guest VMs that use bridged networking.

On older versions of Tart such as 2.10.0, `tart ip --resolver arp <name>` works, but after several hours/days, it stops working and we need to reclone these VMs to get everything back up and working. I'm going to run some longer tests over the weekend with Tart 2.15.0. However, I didn't see anything in the release notes of any recent versions about this, so I suspect the behavior still exists. I'll report back, though.

I'm using macOS Ventura for both the guest and host OS. Is this a known issue?

edigaryev:
Unfortunately that's the reality of using `--net-bridged` and the ARP resolver: we won't know the IP-address of the VM until it sends a broadcast/multicast request that gets accepted by the host and a corresponding IP ↔︎ MAC association added to the ARP table (you can check it with `arp -an`).

This is normally done on startup for macOS guests and almost never on Linux guests, unless [e.g. Samba is installed](https://github.com/cirruslabs/tart/issues/810#issuecomment-2089624865) that generates such requests.

What you can do to work around this is to add a static entry on your local network's DHCP server to always offer a specific IP-address based on the MAC-address of that VM (can be retrieved via `jq -r '.macAddress' ~/.tart/vms/<VM>/config.json`)

nre-ableton:
@edigaryev Thanks for the explanation, that makes a lot of sense. I'll close this issue as "wontfix", since it's not directly a problem in Tart.

nirs:
I could reproduce it:

```
$ tart run --net-bridged en10 server
...
```

(find the ip in the console)

```
% ssh admin@10.100.102.163 ip a   
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host 
       valid_lft forever preferred_lft forever
2: enp0s1: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP group default qlen 1000
    link/ether c2:e1:2c:51:62:20 brd ff:ff:ff:ff:ff:ff
    inet 10.100.102.163/24 metric 100 brd 10.100.102.255 scope global dynamic enp0s1
[...]

PR #1009, opened 2025-01-20
docs(FAQ): document unsupported DHCP client identifiers

edigaryev:
Resolves https://github.com/cirruslabs/softnet/issues/75.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -120,7 +120,9 @@ tart ip --resolver=arp <VM>
 
 This causes the `tart ip` to consult the host's ARP table instead of the `/var/db/dhcpd_leases` file.
 
-Note that this method of resolving the IP heavily relies on the level of VM's talkativeness on the network. This is normally not an issue for macOS VMs, but on Linux VMs you might need to install Samba, which includes a [NetBIOS name server](https://www.samba.org/samba/docs/current/man-html/nmbd.8.html) that exhibits the same behavior as macOS and populates the ARP table of the host OS:
+Note that this method of resolving the IP heavily relies on the level of VM's activity on the network, namely, exchanging ARP requests between the guest and the host.
+
+This is normally not an issue for macOS VMs, but on Linux VMs you might need to install Samba, which includes a [NetBIOS name server](https://www.samba.org/samba/docs/current/man-html/nmbd.8.html) and exhibits the same behavior as macOS, resulting in the population of the ARP table of the host OS:
 
 ```shell
 sudo apt-get install samba
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -81,7 +81,7 @@ And no worries, this file will be re-created on the next `tart run`.
 
 Due to the limitations of the macOS built-in DHCP server, `tart ip` is unable to correctly report the IP addresses for VMs using DHCP client identifiers that are not based on VMs link-layer addresses (MAC addresses).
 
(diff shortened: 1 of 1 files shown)


docs/faq.md
```diff
@@ -77,6 +77,37 @@ sudo rm /var/db/dhcpd_leases
...
+{
+    name=ubuntu
+    ip_address=192.168.64.14
+    hw_address=ff,f1:f5:dd:7f:0:2:0:0:ab:11:cb:fb:30:b0:97:b6:3a:67
+    identifier=ff,f1:f5:dd:7f:0:2:0:0:ab:11:cb:fb:30:b0:97:b6:3a:67
+    lease=0x678e2ce7
+}
+```
+
+Because the macOS built-in DHCP server overwrites the `hw_address` with the `identifier`, it leaves no information about the VM's MAC address to the `tart ip`, which relies on this file to determine the IP address associated with the VM's MAC address.
```
nirs, line 96: It seems like tart ip depends on /var/db/dhcpd_leases, which cannot work for bridged network, since the builtin macOS dhcps server is not the the loop.

It would be nice to add a sentence making this more clear (tart looks up the vm ip address in /var/db/dhcpd_leases...), and add a section about bridged network.

edigaryev: See https://github.com/cirruslabs/tart/pull/1009/commits/98ba3675c006e542aca1b4218afbe867d8d40697.

nirs (commented):
Looks good, but it sounds like we have 2 solutions and users have to choose the right one for the vm. The system can be smarter and pick the right solution (or both) transparently.

docs/faq.md
```diff
@@ -108,6 +110,16 @@ network:
...
 
+## Resolving the VMs IP with bridged networking
+
+When running `tart run` with `--net-bridged`, you need to invoke `tart ip` differently, because the macOS built-in DHCP server won't have any information about the VM's IP-address:
+
+```shell
+tart ip --resolver=arp <VM>
+```
+
+This causes the `tart ip` to consult the host's ARP table instead of the `/var/db/dhcpd_leases` file.
```
nirs, line 121: I have very little experience with querying arp, but in the last time I tried to use it the vm (not a tart vm) got an ip and was accessible, but arp did not report the vm ip.

edigaryev: Probably related to https://github.com/cirruslabs/tart/issues/993#issuecomment-2586640804.

I'll try to document this in this section too.

nirs: This seems to the common case, see
https://github.com/cirruslabs/tart/issues/873#issuecomment-2602347271

edigaryev: Improved in https://github.com/cirruslabs/tart/pull/1009/commits/c21cebfb2d029e6ab45b244e82b87d3935d797b2.

docs/faq.md
```diff
@@ -120,6 +120,12 @@ tart ip --resolver=arp <VM>
 
 This causes the `tart ip` to consult the host's ARP table instead of the `/var/db/dhcpd_leases` file.
 
+Note that this method of resolving the IP heavily relies on the level of VM's talkativeness on the network. This is normally not an issue for macOS VMs, but on Linux VMs you might need to install Samba, which includes a [NetBIOS name server](https://www.samba.org/samba/docs/current/man-html/nmbd.8.html) that exhibits the same behavior as macOS and populates the ARP table of the host OS:
```
nirs, line 123: `talkativeness` sounds wrong to me, maybe it is better to describe the actual behavior (sending the special arp request that makes the host added to arp cache).

edigaryev: Improved in https://github.com/cirruslabs/tart/pull/1009/commits/2e52d9547fc1d7e65770291499b27dd93e57ab59.

PR #1022, opened 2025-02-03
Add support for docker-credential-helpers through environment

nholloh:
## Motivation

Larger organizations running tart often also run their own OCI registry, where VM images are being managed. Different departments within the organization may have limited access to each others images. That is, they may require different sets of credentials to authenticate with a given OCI registry.

Currently, tart only supports a docker-credential-store through a file in `~/.docker/config.json` within the host system. This is not always feasible, as 2 concurrently issued builds in the CI may overwrite each others `config.json` leading to race conditions during OCI authentication.

Alternatively, all authentication with OCI registries can be overridden by setting the `TART_REGISTRY_USERNAME` and `TART_REGISTRY_PASSWORD` variables. However this comes with its own set of issues, as the credentials are not associated with any host. Should a user decide to opt for a publically available image (e.g. [macos-image-templates](https://github.com/cirruslabs/macos-image-templates)) over the internal ones, the default behaviour would have tart send internal credentials to a public OCI registry.

## Current workaround

The current workaround is to execute a script before tart attempts to fetch the image from the OCI registry, where we manually parse a docker-credential-store json string from the environment, select the appropriate credential and set it in tart's `TART_REGISTRY_USERNAME`/`TART_REGISTRY_PASSWORD` environment variables. While this works, it is an unnecessary burden to place on larger organizations, especially if there are existing approaches for similar cases.

## Proposed solution

Similar to [GitLab's approach](https://docs.gitlab.com/ee/ci/docker/authenticate_registry.html#option-3-use-docker_auth_config) for their docker executors, we could allow for setting a docker-credential-store json as environment variable `TART_DOCKER_AUTH_CONFIG`. I chose to prefix the environment variable with `TART_` to remain consistent with existing variables and to avoid conflicts with existing docker setups. tart would then evaluate it in the same way as it currently does the `~/.docker/config.json`. Since the environment variable is specific to the instance of the job being run on the machine and bound to the matching process, concurrently run image fetch operations would not run into the race condition detailed in the motivation above.

## Detailed design

The changes mainly revolve around the `retrieve(host:)` function in the `DockerConfigCredentialsProvider` class. The external interface as `CredentialProvider` remains intact, but the function internally now additionally evaluates the environment. To do so, configurations are now read from the environment and the file system, if available.

```swift
func retrieve(host: String) throws -> (String, String)? {
  let configFromEnvironment = try? self.configFromEnvironment()
  let configFromFileSystem = try? self.configFromFileSystem()

  // ...
}
```

If no configuration is available, we `return nil`.

```swift
guard configFromEnvironment ?? configFromFileSystem != nil else {
  return nil
}
```

The config environment takes precedence over the config from the file system. That is, if authentication is present for a particular host in the environment _and_ in the file system configurations, the values from the environment will be used. Authentication will still succeed even if both configurations are present but the authentication for the desired host can only be found in one of the config objects. Existing logic to prioritize auths over credential helpers has been retained and replicated for the environment.

```swift
if let config = configFromEnvironment {
  if let credentials = config.auths?[host]?.decodeCredentials() {
    return credentials
  }

  if let helperProgram = try config.findCredHelper(host: host) {
    return try executeHelper(binaryName: "docker-credential-\(helperProgram)", host: host)
  }
}

if let config = configFromFileSystem {
  if let credentials = config.auths?[host]?.decodeCredentials() {
    return credentials
  }

  if let helperProgram = try config.findCredHelper(host: host) {
    return try executeHelper(binaryName: "docker-credential-\(helperProgram)", host: host)
  }
}
```

The remaining changes make sure we can inject `ProcessInfo` and `FileManager` as dependencies so that we can properly test the behaviour described above. Tests have been written in `DockerConfigCredentialsProviderTests.swift` using the Swift Testing framework.

## Other changes

I fixed a typo in one of the test file names.


edigaryev:
Hi Niklas 👋

Thanks for your contribution and for taking the time to provide such a details explanation!

I appreciate the effort, but I feel the approach might be a bit more complex than necessary to achieve your goal.

Please take a look at the following alternative:

```patch
diff --git a/Sources/tart/Credentials/DockerConfigCredentialsProvider.swift b/Sources/tart/Credentials/DockerConfigCredentialsProvider.swift
index 6be901f..a50de8e 100644
--- a/Sources/tart/Credentials/DockerConfigCredentialsProvider.swift
+++ b/Sources/tart/Credentials/DockerConfigCredentialsProvider.swift
@@ -2,7 +2,11 @@ import Foundation
 
 class DockerConfigCredentialsProvider: CredentialsProvider {
   func retrieve(host: String) throws -> (String, String)? {
-    let dockerConfigURL = FileManager.default.homeDirectoryForCurrentUser.appendingPathComponent(".docker").appendingPathComponent("config.json")
+    let dockerConfigURL = if let tartDockerConfig = ProcessInfo.processInfo.environment["TART_DOCKER_CONFIG"] {
+      URL(fileURLWithPath: tartDockerConfig)
+    } else {
+      FileManager.default.homeDirectoryForCurrentUser.appendingPathComponent(".docker").appendingPathComponent("config.json")
+    }
     if !FileManager.default.fileExists(atPath: dockerConfigURL.path) {
       return nil
     }
```

The test(s) can be added to [`integration-tests/`](https://github.com/cirruslabs/tart/tree/main/integration-tests), there's no need to mock anything.

nholloh:
The suggestion you provided works, but has some functional key differences. In my design, I chose to allow the user to have _both_, a config from the file system and a config in the environment. While environment takes precedence, if a host is not present in the environment it can still be loaded from the config in the file system. Furthermore, if one of the configs is invalid, the other one will still be evaluated. You can check the unit tests for the details on this.

I also verified my understanding of your approach with the tests I have implemented, but couple of tests now fail.

image

At the end of the day this is a design choice and we can go either way. Regarding the tests, would you like me to remove the unit tests and add integration tests instead or are you fine with keeping the unit tests (and adding integration)?

edigaryev:
>At the end of the day this is a design choice and we can go either way.

I'm all in for a simpler approach.

Also, falling back to a standard `~/.docker/config.json` location when `TART_DOCKER_CONFIG` is explicitly set could be considered magical by some.

>Regarding the tests, would you like me to remove the unit tests and add integration tests instead or are you fine with keeping the unit tests (and adding integration)?

I think we're better off with integration tests since there's no need to introduce unnecessary abstractions to the `CredentialsProvider` implementation this way.

Note that you can also have integration tests in Swift (see `LayerizerTests` for an example).

/cc @fkorotkov

nholloh:
Checking the integration tests I am not sure I understand the approach entirely. How would you

1. pass in the environment variable?
2. make sure that tart authenticates with the right set of credentials against a registry (or the registry runner)?

edigaryev:
> 1. pass in the environment variable?

For that you may need to extend the `Tart.run()`'s signature slightly to accept an additional `env` argument and append it accordingly here:

https://github.com/cirruslabs/tart/blob/f68297097e5f043191b52302fed17f7ff87ce972/integration-tests/tart.py#L24-L32

> 2. make sure that tart authenticates with the right set of credentials against a registry (or the registry runner)?

This can be achieved by simply making sure that `tart pull` and/or `tart push` succeeds against an authenticated registry instance.

We currently use `registry:2`, to make it authenticated for a given test, you might want to configure [native basic auth](https://distribution.github.io/distribution/about/deploying/#native-basic-auth), which again can be made optional in the `DockerRegistry` initializer, so that only specific tests will get the authenticated registry instance: https://github.com/cirruslabs/tart/blob/0b693f6bc9a85475369657869af272b89c6a7cab/integration-tests/docker_registry.py#L10-L12

edigaryev:
> I do like @nholloh's `TART_DOCKER_AUTH_CONFIG` name more though. Just `TART_DOCKER_CONFIG` seems not clear enough, Tart only uses this config for authentication.

The issue I see with this naming is that `~/.docker/config.json` is a general configuration file and it is not auth-specific.

The other point is that [`DOCKER_CONFIG` is an well-known environment variable used by Docker](https://docs.docker.com/reference/cli/docker/#environment-variables).

nholloh:
> The issue I see with this naming is that `~/.docker/config.json` is a general configuration file and it is not auth-specific.
> 
> The other point is that [`DOCKER_CONFIG` is an well-known environment variable used by Docker](https://docs.docker.com/reference/cli/docker/#environment-variables).

Since this change is a proposal in the context of CI, specifically due to the issues we're facing on GitLab, I see value in following GitLab's naming scheme. 

Additionally, the `DOCKER_CONFIG` variable you reference is described as:

> The location of your client configuration files.

While I see your point that the configuration may contain more information than just authentication, the new environment variable I am proposing is carrying the file's _contents_ rather than a _path to the file_. Using the same naming scheme for entirely different content could lead to confusion around its behaviour. Providing a file path instead of the content would not resolve the issue mentioned above without further workarounds. Therefore, I still think explicitly choosing a different name makes sense.

How about `TART_DOCKER_CONFIG_JSON`?

edigaryev:
> Since this change is a proposal in the context of CI, specifically due to the issues we're facing on GitLab, I see value in following GitLab's naming scheme.

This seems like a rather stretched assumption because Tart is used in various environments, not just GitLab, and there our users may find this environment variable useful as well.

>I am proposing is carrying the file's contents rather than a path to the file

>Providing a file path instead of the content would not resolve the issue mentioned above without further workarounds.

Just to clarify, is there a reason this approach doesn't this work for you?

```yaml
script:
  - echo "$DOCKER_CONFIG_BASE64" | base64 -d > ./config.json
  - export TART_DOCKER_CONFIG=./config.json
  - tart ...
```

It's used in GitLab for `KUBECONFIG`, `GOOGLE_APPLICATION_CREDENTIALS` and other environment variables that contain a path to a file.

The file will be created in the project's directory and shouldn't cause any conflicts with other jobs.

nholloh:
Coming from iOS development, I don't really know my way around the different tools used in infra and hosting. That is, if it really is GitLab who deviate from the convention to pass file paths into a tool as environment variable for the OCI registry authentication, then I am completely with you. In this case we should absolutely pass the file path as env var.

To make things align with GitLab's way of providing credentials, maybe we could add the feature to the prepare command of the [gitlab-tart-executor](https://github.com/cirruslabs/gitlab-tart-executor/blob/main/internal/commands/prepare/prepare.go). It could then take a `TART_DOCKER_AUTH_CONFIG` (config file content), writes it to a temporary file and passes the path to that file as `TART_DOCKER_CONFIG` to tart? This would give GitLab users a familiar tool to work with while abstracting away the added complexity of writing the content to a temporary file.

edigaryev:
@nholloh that sounds like good trade-off to me, especially since we'd still need to change the GitLab Tart Executor to support this because GitLab Runner prepends `CUSTOM_ENV_` to CI/CD environment variables.

See for example how we do this for `TART_REGISTRY_{USERNAME,PASSWORD}`: https://github.com/cirruslabs/gitlab-tart-executor/blob/e03add8340a51aa1e51d9e5269b42951a1a3e713/internal/commands/prepare/prepare.go#L279-L307.

However, I still think that a more appropriate name for the environment variable should be a suffix to `TART_DOCKER_CONFIG` to make it more discoverable. For example, `TART_DOCKER_CONFIG_CONTENTS`, `TART_DOCKER_CONFIG_JSON` or `TART_DOCKER_CONFIG_BASE64`.

edigaryev:
> If, however, the `TART_DOCKER_CONFIG` env var was set explicitly, I would expect tart to fail with an error and tell me the file was not found.

Sounds reasonable to me, as long as the user will get a clear error message explaining the error.

nholloh:
Done, the changes are implemented. Feel free to take a look. The error message now reads:

```
Error: Registry authentication failed. Could not find docker configuration at '/tmp/abc'. The file doesn’t exist.
```

It uses the standard apple keys and error codes for the FileNotFound error.

Fred78290:
Just a remark!
Don't think that docker registry are only authentified with user/password. So it means that often docker-credential-helpers is needed and .docker/config.json is not enough.

As example the authent to AWS ECR, using AWS secret/access key or ephemeral token. Azure too. 

As ref: https://aws.amazon.com/fr/blogs/compute/authenticating-amazon-ecr-repositories-for-docker-cli-with-credential-helper/

Fred78290:
> Would this not be covered through the existing implementation?
> 
> ```swift
> private func retrieveCredentials(for host: String, from config: DockerConfig) throws -> (String, String)? {
>     if let credentials = config.auths?[host]?.decodeCredentials() {
>       return credentials
>     }
> 
>     if let helperProgram = try config.findCredHelper(host: host) {
>       return try executeHelper(binaryName: "docker-credential-\(helperProgram)", host: host)
>     }
> 
>     return nil
>   }
> ```
At minima with your implementation, on aws or azure you must create a temporary token to authent on private registry in your script such aws aws ecr-login....

But it not enough if you want integrate docker credentials helper mechanism with existing plugin.

Take a look at:

https://github.com/docker/docker-credential-helpers
https://github.com/awslabs/amazon-ecr-credential-helper

Sources/tart/Credentials/DockerConfigCredentialsProvider.swift
```diff
@@ -1,16 +1,54 @@
 import Foundation
 
 class DockerConfigCredentialsProvider: CredentialsProvider {
+
   func retrieve(host: String) throws -> (String, String)? {
```
edigaryev, line 5: Sorry, but this is too much indirection for such a simple function. We now have four new functions containing only 5 lines of code on average, and they're only being called from a single place, thus no code re-use at all.

I think we could get away with just a single new function:

```swift
private func dockerConfig() throws -> Data? { ... }
```

That will either read from `TART_DOCKER_CONFIG` or from the default location, and call it from the `retrieve()` accordingly.

edigaryev (changes requested):
Please, try make the smallest amount of changes necessary to get your functionality across.

This will save both your (writing the code) and our time (reviewing it).

PR #1046, opened 2025-03-24
Enable clipboard sharing on macOS too

edigaryev:
And document which packages need to be installed on these operating systems.

Resolves https://github.com/cirruslabs/tart/issues/14.

### DIFF
--- a/Sources/tart/Commands/Run.swift
+++ b/Sources/tart/Commands/Run.swift
@@ -92,7 +92,7 @@ struct Run: AsyncParsableCommand {
 
   @Flag(help: ArgumentHelp(
     "Disable clipboard sharing between host and guest.",
-    discussion: "Clipboard sharing requires spice-vdagent package on Linux and https://github.com/utmapp/vd_agent on macOS."))
+    discussion: "Clipboard sharing requires spice-vdagent package on Linux and https://github.com/cirruslabs/tart-guest-agent on macOS."))
   var noClipboard: Bool = false
 
   #if arch(arm64)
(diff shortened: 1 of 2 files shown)


Uzaaft:
@edigaryev Out of curiousity:
Does this work with wayland?

edigaryev:
> Does this work with wayland?

Doesn't seem so: https://gitlab.freedesktop.org/spice/linux/vd_agent/-/issues/26.

By the way, which Linux distributions ship with Wayland enabled by default?

amalchuk:
@edigaryev Any plans to fork/contribute [vd_agent](https://github.com/utmapp/vd_agent) for shipping a binaries as it was done in [softnet](https://github.com/cirruslabs/softnet)?

Uzaaft:
Seems like https://wayland.app/protocols/ext-data-control-v1 is what some people/solutions use to enable clipboard sharing.

edigaryev:
> @edigaryev Any plans to fork/contribute [vd_agent](https://github.com/utmapp/vd_agent) for shipping a binaries as it was done in [softnet](https://github.com/cirruslabs/softnet)?

We've implemented clipboard sharing functionality in our [Tart Guest Agent](https://github.com/cirruslabs/tart-guest-agent), which is compatible with [SPICE vdagent protocol](https://www.spice-space.org/agent-protocol.html).

This agent is now present in all non-vanilla Cirrus Labs macOS VM images, and after a week or so once the new version gets baked-in, the clipboard sharing should start to work on most of our macOS images out-of-the box.

PR #1066, opened 2025-05-06
Fix unescaped commas in generated ArgumentParser completions

edigaryev:
See https://github.com/apple/swift-argument-parser/issues/761 for more details.

Resolves https://github.com/cirruslabs/tart/issues/1063.


edigaryev:
I've also improved completion hints in https://github.com/cirruslabs/tart/pull/1066/commits/0a346cfa539a6f5a248fa1728a4528c9a96741c3, so that `tart run --disk`, `tart run --dir` and others actually complete file/directory/local VM directory names.

PR #1074, opened 2025-05-16
Introduce "tart exec" command as an alternative to SSH

edigaryev:
See https://github.com/cirruslabs/tart-guest-agent/pull/10.


edigaryev:
@fkorotkov PTAL once more, NIO async/await primitives simplified things very nicely.

PR #1086, opened 2025-06-05
 Allow VM window to be resized smaller than its configured resolution

Gavinkaa:
Allow the VM window to be resized below its configured display resolution by removing the `minWidth`/`minHeight` constraints in `Run.swift`. This lets users on smaller host screens shrink the window for better multitasking.

See issue #1087

### DIFF
--- a/Sources/tart/Commands/Run.swift
+++ b/Sources/tart/Commands/Run.swift
@@ -745,10 +745,8 @@ struct MainApp: App {
           }
         }
       }.frame(
-        minWidth: CGFloat(vm!.config.display.width),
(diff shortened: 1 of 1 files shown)


edigaryev:
I've tried this change locally and I'm afraid that this is not a good UX because it's almost impossible to get back to the original dimensions after you fiddle with the Tart window size.

Whereas without this change, you can just resize the window to it's smallest dimensions.

>This lets users on smaller host screens shrink the window for better multitasking.

I think this can be achieved simply by modifying your VM as follows:

```shell
tart set --display 1x1 <VM name>
```

PR #1094, opened 2025-06-14
feat: Add disk image format selection with ASIF support

fkorotkov:
Used http://augmentcode.com/ to generate the following in 3 prompts. See each commit.

This adds support ASIF (Apple Sparse Image Format) provides better performance on macOS 26 Tahoe and later systems while maintaining compatibility with existing RAW disk workflows.

In addition each VM config now has a metadata about the format and `tart set --disk-size N` will use `diskutils` to resize which is reflected in the VM (no need to SSH and run the extra step inside of the guest).


Tests/TartTests/DiskImageFormatTests.swift
```diff
@@ -0,0 +1,77 @@
...
+    }
+  }
+
+  func testFormatFromString() throws {
+    XCTAssertEqual(DiskImageFormat(rawValue: "raw"), .raw)
+    XCTAssertEqual(DiskImageFormat(rawValue: "asif"), .asif)
+    XCTAssertNil(DiskImageFormat(rawValue: "invalid"))
+  }
+
+  func testFormatArgument() throws {
```
edigaryev, line 25: 1. This test should probably be named `testCaseInsensitivity()` or something like that, `testFormatArgument()` is too vague.

2. There's no need to have `"raw"` and `"invalid"` test cases here as they're already tested in `testFormatFromString ()`.

Tests/TartTests/DiskImageFormatTests.swift
```diff
@@ -0,0 +1,77 @@
...
+    XCTAssertEqual(DiskImageFormat(argument: "raw"), .raw)
+    XCTAssertEqual(DiskImageFormat(argument: "ASIF"), .asif) // case insensitive
+    XCTAssertEqual(DiskImageFormat(argument: "Raw"), .raw)  // case insensitive
+    XCTAssertNil(DiskImageFormat(argument: "invalid"))
+  }
+
+  func testFormatDescriptions() throws {
```
edigaryev, line 32: Do we really need this test? It doesn't look very useful.

Sources/tart/Commands/Create.swift
```diff
@@ -28,6 +31,16 @@ struct Create: AsyncParsableCommand {
         throw ValidationError("Only Linux VMs are supported on Intel!")
       }
     #endif
+
+    // Validate disk format support
```
edigaryev, line 35: 1. Let's have the same validation in `tart run` since only macOS Tahoe can work with ASIF.

2. Since `diskFormat.isSupported` is the same thing as `diskFormat.canCreate`, we can probably remove the latter.

Sources/tart/Commands/Create.swift
```diff
@@ -19,6 +19,9 @@ struct Create: AsyncParsableCommand {
   @Option(help: ArgumentHelp("Disk size in GB"))
   var diskSize: UInt16 = 50
 
+  @Option(help: ArgumentHelp("Disk image format: raw (default) or asif", discussion: "ASIF format provides better performance but requires macOS 26 Tahoe or later"))
```
edigaryev, line 22: `ArgumentParser` already shows the possible values and the default value:

```suggestion
  @Option(help: ArgumentHelp("Disk image format", discussion: "ASIF format provides better performance but requires macOS 26 Tahoe or later"))
```

Sources/tart/DiskImageFormat.swift
```diff
@@ -0,0 +1,89 @@
...
+import ArgumentParser
+
+enum DiskImageFormat: String, CaseIterable, Codable {
+  case raw = "raw"
+  case asif = "asif"
+
+  var displayName: String {
+    switch self {
+    case .raw:
+      return "RAW (UDIF read-write)"
```
edigaryev, line 11: I think RAW is unrelated to UDIF:

```suggestion
      return "RAW"
```

Sources/tart/DiskImageFormat.swift
```diff
@@ -0,0 +1,89 @@
...
+      return "ASIF (Apple Sparse Image Format)"
+    }
+  }
+
+  var description: String {
+    switch self {
+    case .raw:
+      return "Traditional disk image format, compatible with all macOS versions"
+    case .asif:
+      return "High-performance sparse disk image format, requires macOS 15+ (creation requires macOS 26+)"
```
edigaryev, line 22:
```suggestion
      return "High-performance sparse disk image format, requires macOS 26 or newer"
```

Also, not sure if `description` is even used at all (dead code), we can just probably remove it altogether.

Sources/tart/VM.swift
```diff
@@ -383,13 +385,22 @@ class VM: NSObject, VZVirtualMachineDelegate, ObservableObject {
     }
 
     // Storage
-    var attachment = try VZDiskImageStorageDeviceAttachment(
-      url: diskURL,
-      readOnly: false,
+    let defaultCachingMode: VZDiskImageCachingMode
+    if vmConfig.diskFormat == .asif {
+      // For ASIF disks, use .uncached mode to avoid compatibility issues
+      defaultCachingMode = .uncached
```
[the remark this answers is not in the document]
edigaryev: Changed this to `.cached` locally and it works just fine on `macOS 26.0 (25A5279m)`.

Perhaps we can just leave it as is and collect user reports.

Sources/tart/VMDirectory.swift
```diff
@@ -161,6 +181,175 @@ struct VMDirectory: Prunable {
     try diskFileHandle.close()
   }
 
+  private func resizeASIFDisk(_ sizeGB: UInt16) throws {
```
edigaryev, line 184: It would be logical to move this function and functions below to `DiskImageFormat` instead now that we have it.

Sources/tart/VMDirectory.swift
```diff
@@ -161,6 +181,175 @@ struct VMDirectory: Prunable {
...
+      if infoProcess.terminationStatus != 0 {
+        let output = String(data: infoData, encoding: .utf8) ?? "Unknown error"
+        throw RuntimeError.FailedToResizeDisk("Failed to get ASIF disk info: \(output)")
+      }
+
+      // Parse the plist to get current size
+      do {
+        let plist = try PropertyListSerialization.propertyList(from: infoData, options: [], format: nil) as? [String: Any]
+
+        if let plist = plist {
```
edigaryev, line 213: Too much nesting, please use [`guard let`](https://www.hackingwithswift.com/quick-start/understanding-swift/when-to-use-guard-let-rather-than-if-let) to fix this.

Sources/tart/VMDirectory.swift
```diff
@@ -161,6 +181,175 @@ struct VMDirectory: Prunable {
...
+          // Look for size information in the "Size Info" dictionary
+          var currentSizeBytes: UInt64?
+
+          // For ASIF images, the size is in the "Size Info" dictionary under "Total Bytes"
+          if let sizeInfo = plist["Size Info"] as? [String: Any],
```
edigaryev, line 218: This could can be simplified and made more type-safe by using [`PropertyListDecoder`](https://developer.apple.com/documentation/foundation/propertylistdecoder) instead.

Sources/tart/VM.swift
```diff
@@ -383,13 +385,19 @@ class VM: NSObject, VZVirtualMachineDelegate, ObservableObject {
     }
 
     // Storage
-    var attachment = try VZDiskImageStorageDeviceAttachment(
```
edigaryev, line 386: Would it be possible to just revert this part of code?

It doesn't need any changes.

Sources/tart/OCI/Manifest.swift
```diff
@@ -13,6 +13,7 @@ let nvramMediaType = "application/vnd.cirruslabs.tart.nvram.v1"
 // Manifest annotations
 let uncompressedDiskSizeAnnotation = "org.cirruslabs.tart.uncompressed-disk-size"
 let uploadTimeAnnotation = "org.cirruslabs.tart.upload-time"
+let diskFormatLabelAnnotation = "org.cirruslabs.tart.disk.format"
```
edigaryev, line 16: Not: this is a label, not an annotation.

So it should probably go to `// Manifest labels` section and be named `diskFormatLabel`.

PR #1102, opened 2025-07-07
feat: prioritize pruning of old SHA when pulling updated tags

fkorotkov:
When pulling a new version of a tagged image (e.g., ghcr.io/cirruslabs/macos-runner:sonoma), set the access date of the previous SHA to epoch time (1970-01-01). This ensures that the old SHA will be prioritized for pruning, even if it was accessed more recently than other cached images.

This helps manage disk space more efficiently by automatically cleaning up superseded versions of frequently-updated tagged images.

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/Sources/tart/VMStorageOCI.swift
+++ b/Sources/tart/VMStorageOCI.swift
@@ -27,12 +27,12 @@ class VMStorageOCI: PrunableStorage {
     return digest
   }
 
-  func open(_ name: RemoteName) throws -> VMDirectory {
+  func open(_ name: RemoteName, _ accessDate: Date = Date()) throws -> VMDirectory {
     let vmDir = VMDirectory(baseURL: vmURL(name))
 
     try vmDir.validate(userFriendlyName: name.description)
 
-    try vmDir.baseURL.updateAccessDate()
+    try vmDir.baseURL.updateAccessDate(accessDate)
 
     return vmDir
   }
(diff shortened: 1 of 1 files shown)


Sources/tart/VMStorageOCI.swift
```diff
@@ -180,6 +180,17 @@ class VMStorageOCI: PrunableStorage {
       let transaction = SentrySDK.startTransaction(name: name.description, operation: "pull", bindToScope: true)
       let tmpVMDir = try VMDirectory.temporaryDeterministic(key: name.description)
 
+      // Check if there's an existing symlink pointing to a different digest
+      if exists(name) {
```
edigaryev, line 184: What do you think about simplifying this down to:

```swift
// Open an existing VM directory corresponding to this name, if any,
// marking it as outdated to speed up the garbage collection process
_ = try? open(name, Date(timeIntervalSince1970: 0))
```

The `open()` will need to be adjusted as follows:

```patch
diff --git a/Sources/tart/VMStorageOCI.swift b/Sources/tart/VMStorageOCI.swift
index 764eb1e..f71c547 100644
--- a/Sources/tart/VMStorageOCI.swift
+++ b/Sources/tart/VMStorageOCI.swift
@@ -27,12 +27,12 @@ class VMStorageOCI: PrunableStorage {
     return digest
   }
 
-  func open(_ name: RemoteName) throws -> VMDirectory {
+  func open(_ name: RemoteName, _ accessDate: Date = Date()) throws -> VMDirectory {
     let vmDir = VMDirectory(baseURL: vmURL(name))
 
     try vmDir.validate(userFriendlyName: name.description)
 
-    try vmDir.baseURL.updateAccessDate()
+    try vmDir.baseURL.updateAccessDate(accessDate)
 
     return vmDir
   }
```

edigaryev (approved):

Issue #1100, opened 2025-07-06
tart exec -i hangs when input is piped via stdin


therealpxc:
I tried using `tart exec -i` to run a script without transferring it over to a VM's filesystem, and it works, but once the entirety of the script has been read, I'm not properly returned to my terminal prompt on the host machine. I have to type `^C` to get out.

Compare this to the behavior of `docker`:

```
❯ cat trivial.sh 
echo hi

❯ docker run -i bash < trivial.sh
hi

❯ tart exec -i sandbox sh < trivial.sh
hi
^CError: CancellationError()

❯ 
```

In case anyone else is hitting this, the commonsense workaround does work, at least for short scripts:

```
❯ tart exec -i sandbox sh -c "$(cat trivial.sh)"
hi

❯ 
```

but I think that will choke on long scripts because of OS limits on the number of CLI args on the guest side.

therealpxc:
I did a bit more playing around, and Interestingly, it seems that I don't have this issue with pipes from other commands, only the `<` operator built into shells. So using a "useless use of cat" is a better workaround than the one I originally devised.

PR #1106, opened 2025-07-10
tart exec: handle input redirection of regular files

edigaryev:
Resolves https://github.com/cirruslabs/tart/issues/1100.


therealpxc:
Thanks!

Issue #1112, opened 2025-08-02
Error after moving TART_HOME to external drive


matthewruzzi:
I copied the `.tart` folder to an external drive and added `export TART_HOME="/Volumes/2TB\ Envoy\ Express\ 2/.tart"` to my `.zprofile`. I now get an error message whenever I try to use tart. 

```console
matthew@Matthews-Mac-mini ~ % tart
Failed to perform garbage collection!
Error Domain=NSCocoaErrorDomain Code=513 "You don’t have permission to save the file “cache” in the folder “.tart”." UserInfo={NSFilePath=/Volumes/2TB\ Envoy\ Express\ 2/.tart/cache, NSURL=file:///Volumes/2TB%5C%20Envoy%5C%20Express%5C%202/.tart/cache, NSUnderlyingError=0x600002ca4810 {Error Domain=NSPOSIXErrorDomain Code=13 "Permission denied"}}
USAGE: tart <subcommand>

OPTIONS:
  --version               Show the version.
  -h, --help              Show help information.

SUBCOMMANDS:
  create                  Create a VM
  clone                   Clone a VM
  run                     Run a VM
  set                     Modify VM's configuration
  get                     Get a VM's configuration
  list                    List created VMs
  login                   Login to a registry
  logout                  Logout from a registry
  ip                      Get VM's IP address
  exec                    Execute a command in a running VM
  pull                    Pull a VM from a registry
  push                    Push a VM to a registry
  import                  Import VM from a compressed .tvm file
  export                  Export VM to a compressed .tvm file
  prune                   Prune OCI and IPSW caches or local VMs
  rename                  Rename a local VM
  stop                    Stop a VM
  delete                  Delete a VM
  suspend                 Suspend a VM

  See 'tart help <subcommand>' for detailed help.
matthew@Matthews-Mac-mini ~ % tart run ubuntu

tart/VMStorageLocal.swift:4: Fatal error: 'try!' expression unexpectedly raised an error: Error Domain=NSCocoaErrorDomain Code=513 "You don’t have permission to save the file “cache” in the folder “.tart”." UserInfo={NSFilePath=/Volumes/2TB\ Envoy\ Express\ 2/.tart/cache, NSURL=file:///Volumes/2TB%5C%20Envoy%5C%20Express%5C%202/.tart/cache, NSUnderlyingError=0x60000320cd50 {Error Domain=NSPOSIXErrorDomain Code=13 "Permission denied"}}
[...]

PR #1138, opened 2025-09-24
Validate custom TART_HOME and provide a human-friendly error message

edigaryev:
See https://github.com/cirruslabs/tart/issues/1112#issuecomment-3235462418.


edigaryev:
Seems like `deletingLastPathComponent ()` adds `../../../` infinitely in some cases, fixing this in https://github.com/cirruslabs/tart/pull/1138/commits/8c804c72387bfc2e3d3e0964e5402267da0a1a58.

PR #1139, opened 2025-09-27
Add NoGraphics field to tart get command

webcoyote:
I want to detect whether a VM is running with a window or without for [ClodPod](https://github.com/webcoyote/clodpod), a VM sandbox for AI agents, so I added a feature to `tart`:

- Added `NoGraphics` field to the `tart get` command output
- Returns `false` when window is present, `true` when window not present, and `null` when VM is not running
- Includes integration tests to verify the functionality

Glad to make changes; I agonized over whether the field should be `NoGraphics: true/false` or `Graphics: false/true`.

NOTE: this is an awkward feature to test because in CI mode (like GitHub actions), the CI environment never creates windows, so I can only test for "--no-graphics" and "not running".

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,6 @@ dist/
 
 # mkdocs-material
 site
+
+# Python cache files
+*.pyc
--- a/Sources/tart/Commands/Get.swift
+++ b/Sources/tart/Commands/Get.swift
@@ -1,5 +1,7 @@
 import ArgumentParser
 import Foundation
+import CoreGraphics
+import Darwin
 
 fileprivate struct VMInfo: Encodable {
   let OS: OS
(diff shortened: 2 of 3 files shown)


edigaryev:
> I want to detect whether a VM is running with a window or without for [ClodPod](https://github.com/webcoyote/clodpod), a VM sandbox for AI agents

Have you considered storing this state in your software instead?

As far as I understand, it is ClodPod that [invokes `tart run` with `--no-graphics`](https://github.com/webcoyote/clodpod/blob/c20db9732d147458b105e0262860e8937f1e40cd/clod#L135) conditionally in the first place.

Since [you also use SQLite](https://github.com/webcoyote/clodpod/blob/c20db9732d147458b105e0262860e8937f1e40cd/clod#L142-L144), you can just store an additional `no_graphics` boolean column and set it at the time VM creation to achieve the same result.

webcoyote:
That's a reasonable solution, though whatever cached state ClodPod has might not accurately reflect the actual state of the virtual machine.

PR #1140, opened 2025-09-29
Avoid duplicate progress updates in CI logs

stmitt:
We run tart inside a GitHub Action and noticed that progress updates flood the log because the non-interactive console prints every updateLastLine call, even when the percentage hasn’t actually changed. 

With this change the ProgressObserver remembers the last rendered percentage and skips redundant updates, so the action log now shows each percent only once while keeping interactive behavior unchanged.

### DIFF
--- a/Sources/tart/Logging/ProgressObserver.swift
+++ b/Sources/tart/Logging/ProgressObserver.swift
@@ -20,7 +20,7 @@ public class ProgressObserver: NSObject {
         self.lastTimeUpdated = currentTime
         let line = ProgressObserver.lineToRender(self.progressToObserve)
         // Skip identical renders so non-interactive logs only see new percent values.
-        guard line != self.lastRenderedLine else {
+        if line == self.lastRenderedLine {
           return
         }
 


Sources/tart/Logging/ProgressObserver.swift
```diff
@@ -4,18 +4,28 @@ public class ProgressObserver: NSObject {
...
     observation = observe(\.progressToObserve.fractionCompleted) { progress, _ in
       let currentTime = Date.now
       if self.progressToObserve.isFinished || currentTime.timeIntervalSince(self.lastTimeUpdated) >= 1.0 {
         self.lastTimeUpdated = currentTime
-        renderer.updateLastLine(ProgressObserver.lineToRender(self.progressToObserve))
+        let line = ProgressObserver.lineToRender(self.progressToObserve)
+        // Skip identical renders so non-interactive logs only see new percent values.
+        guard line != self.lastRenderedLine else {
+          return
+        }
```
edigaryev, line 25: It feels like this could be simplified:

```suggestion
        if line == self.lastRenderedLine {
          return
        }
```

Otherwise LGTM.

edigaryev (approved):

Issue #1150, opened 2025-10-14
Display Sizing Confusion


brianmichel:
It seems like there are some interconnected options when trying to set the display size of a virtual machine that are undocumented within Tart. 

When trying to set the display size, naively I tried `tart set <vm-name> --display 1280x960`, and then `tart run <vm-name>`. This boots the VM in GUI mode, but depending on the size of the window (which has perhaps has had it's frame auto-restored by macOS) your VM will not be 1280x960. It seems if if you want a locked size on startup when running in GUI mode for a VM, you have to also pass the `--no-display-refit` option to ensure things don't get resized. 

Interestingly, if I'm not running in GUI mode `tart run <vm-name> --no-graphics --vnc` the passed in display size seems to work just fine. I assume this is because there is no GUI window being created which would trigger whatever implied display retrofitting logic is tied to that window size.

As a user, I would assume that if I pass a value for `--display` that I'm implying `--no-display-refit`. If I don't pass a `--display` option as part of `set` I would assume that the GUI window would control the size of things.

I'd like to confirm this is the intended behavior, and if so, make the suggestion consider the above statement on how these options should work, or perhaps just need to be better documented.

I'm using Tart version `2.29.0`

brianmichel:
Alternatively, is there a way to set the default resolution of a VM? I can't completely tell as it seems like the display values are only used to set the dimensions of the SwiftUI window that gets created if you're running in GUI mode.

PR #1151, opened 2025-10-15
Allow Specification Of Display Unit Type

brianmichel:
Allow the specification of the backing display unit type (pixels or points) to provide greater control of the screen dimensions for CI/CD environment purposes. 

This should also ensure backward compatibility with existing configurations that do not have this config value present and default it to `point` as that has been used as the default.


brianmichel:
I think this fixes https://github.com/cirruslabs/tart/issues/1150 by ensuring we don't do any kind of relative resolution changes if someone has explicitly opted out with `--no-display-refit`. 

I did wonder if a better value to determine whether or not relative resizing needed to take place was the `--no-graphics` flag instead since I think the same mental model could apply if people are using that flag.

brianmichel:
@edigaryev lmk if this is roughly the direction you were thinking and I'll get the tests passing

PR #1155, opened 2025-10-21
tart set: support optional "pt" and "px" units for "--display" argument

edigaryev:
A simpler alternative to https://github.com/cirruslabs/tart/pull/1151, without introducing a new command-line argument.

Resolves https://github.com/cirruslabs/tart/issues/1150, closes https://github.com/cirruslabs/tart/pull/1151.

### DIFF
--- a/Sources/tart/Commands/Set.swift
+++ b/Sources/tart/Commands/Set.swift
@@ -56,6 +56,7 @@ struct Set: AsyncParsableCommand {
       if (display.height > 0) {
         vmConfig.display.height = display.height
       }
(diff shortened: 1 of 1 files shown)


edigaryev:
> The new parser correctly recognizes `pt`/`px` suffixes, but `run()` only copies `width` and `height` into the VM config and drops the parsed `unit`.

Thanks, fixed in https://github.com/cirruslabs/tart/pull/1155/commits/72b58b507859252f6d354dc874615b3f35ad8627.

Issue #1175, opened 2026-01-19
Completion script causes shell errors before first run


CamJN:
the completion script installed by the homebrew formula:

tries to create `~/Library/Caches/org.cirruslabs.tart/io.sentry` but `~/Library/Caches/org.cirruslabs.tart` doesn't exist at that point and therefore it causes errors to be printed to the terminal when the completion script loads:

```
-bash: Cant create base path to store launch profile config file: Error Domain=SentryErrorDomain Code=108 "Failed to create the directory at path /Users/camdennarzt/Library/Caches/org.cirruslabs.tart/io.sentry." UserInfo={NSLocalizedDescription=Failed to create the directory at path /Users/camdennarzt/Library/Caches/org.cirruslabs.tart/io.sentry., NSUnderlyingError=0xb4b048060 {Error Domain=NSCocoaErrorDomain Code=513 "You don’t have permission to save the file “io.sentry” in the folder “org.cirruslabs.tart”." UserInfo={NSFilePath=/Users/camdennarzt/Library/Caches/org.cirruslabs.tart/io.sentry, NSURL=file:///Users/camdennarzt/Library/Caches/org.cirruslabs.tart/io.sentry, NSUnderlyingError=0xb4b048000 {Error Domain=NSPOSIXErrorDomain Code=1 "Operation not permitted"}}}}

#!/bin/bash

__tart_cursor_index_in_current_word() {
    local remaining="${COMP_LINE}"

    local word
    for word in "${COMP_WORDS[@]::COMP_CWORD}"; do
        remaining="${remaining##*([[:space:]])"${word}"*([[:space:]])}"
    done

    local -ir index="$((COMP_POINT - ${#COMP_LINE} + ${#remaining}))"
    if [[ "${index}" -le 0 ]]; then
        printf 0
    else
        printf %s "${index}"
    fi
}

# positional arguments:
#
# - 1: the current (sub)commands: File name too long
```

CamJN:
Looks like a regression of https://github.com/cirruslabs/tart/issues/1163
```
$ sw_vers -productVersion
26.2
$ tart --version
2.30.1
$ brew --version
Homebrew 5.0.10-65-ga36eea6
$ bash --version
GNU bash, version 5.3.9(1)-release (aarch64-apple-darwin25.1.0)
```

PR #1179, opened 2026-01-22
Switch to OpenTelemetry

edigaryev:
Tested locally with [`otel-tui`](https://github.com/ymtdzzz/otel-tui).

See https://github.com/cirruslabs/tart/issues/1175#issuecomment-3778999924.


fkorotkov (commented):
Can `otel-tui` be used by a coding agent to test similar things? Is there non-interractive mode? Kind of run in background and then query that it saw things. Maybe run as background with `--debug-log ` (`/experimental` in codex to enable background) and then check the log?

Overall LGTM

edigaryev:
> Wish to see the skill first.

I've just tried writing a skill for this, and I'm not sure if it's the best approach to solve the problem.

A simple integration test (which could also be written by the same model) would:

* yield reproducible results every time, without hallucinations
* run much faster
* require minimal (allow the sandbox to execute the test suite) to no user input
* would not waste tokens on each invocation

I've used the following prompt:

```
Ensure that "./scripts/run-signed.sh list" produces OpenTelemetry data using $opentelemetry-collector-debug
```

And the following skill: [`opentelemetry-collector-debug/SKILL.md`](https://github.com/user-attachments/files/24807072/SKILL.md).

PR #1204, opened 2026-02-25
Remove disk v1 support

fkorotkov:


edigaryev (dismissed):
LGTM, but let's add a nicer error when `application/vnd.cirruslabs.tart.disk.v1` layers are present.

Something like:

```swift
throw RuntimeError.Generic("Pulling OCI images with legacy disk media type \"application/vnd.cirruslabs.tart.disk.v1\" is no longer supported, please re-push the image using a current Tart version")
```

PR #1206, opened 2026-02-27
Replace Sequoia with Tahoe in Quick Start's first few instructions & add Tahoe to images available

minihoot:


fkorotkov (approved):
Thank you!

PR #1238, opened 2026-06-05
Relicense under FSL-1.1-ALv2

fkorotkov-oai:
We will migrate CI in a follow-up after this change and before the next release.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,45 +1,105 @@
-Fair Source License, version 0.9
+# Functional Source License, Version 1.1, ALv2 Future License
 
-Copyright (C) 2023 Cirrus Labs, Inc.
+## Abbreviation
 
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -1,45 +1,105 @@
...
-Software from you.
+# Functional Source License, Version 1.1, ALv2 Future License
+
+## Abbreviation
+
+FSL-1.1-ALv2
+
+## Notice
+
+Copyright 2026 OpenAI
```
edi-oai: Should this cover the project's entire lifetime, i.e. `2022–2026`, rather than just `2026`?

fkorotkov-oai: Good catch, updated.

edi-oai (approved):

PR #1260, opened 2026-06-09
Don't run the AppKit run loop nested in Swift's async main

torarnv:
When built against the macOS 27 (Xcode 27, Swift 6.4) SDK, "tart run" brings up the VM window but the guest never boots.

Swift's asynchronous main() entry point implicitly starts an executor that owns the main thread, and as of Swift 6.4 that executor is no longer backed by the Dispatch main queue. Running an AppKit/SwiftUI run loop nested inside it via MainApp.main() leaves the main run loop unable to drain Swift tasks or DispatchQueue.main, so the task that starts the VM is never scheduled, even though the window itself (driven directly by AppKit during launch) still appears.

We now keep Root.main() synchronous, so that a command driving a run loop can own the main thread at the top level, exactly like a plain SwiftUI app. With AppKit owning the loop again, MainActor tasks and the Dispatch main queue drain as before. Such commands opt in through a new MainThreadCommand protocol; everything else keeps running asynchronously via a detached task and dispatchMain().

Verified that the guest boots again, and that Ctrl+C still stops the VM gracefully.

### DIFF
--- a/Sources/tart/Commands/Run.swift
+++ b/Sources/tart/Commands/Run.swift
@@ -360,7 +360,7 @@ struct Run: AsyncParsableCommand {
   }
 
   @MainActor
-  func run() async throws {
+  func runOnMainThread() throws {
     let localStorage = try VMStorageLocal()
     let vmDir = try localStorage.open(name)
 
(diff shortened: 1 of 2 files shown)


torarnv:
@fkorotkov-oai @edigaryev

fkorotkov-oai (approved):
Thank you!

PR #1263, opened 2026-06-09
Add --provisioning-opts flag to provision macOS guests on first boot

torarnv:
Exposes Apple's macOS 27 guest provisioning API
(VZMacGuestProvisioningOptions) so a macOS guest can be set up automatically on the first boot after restore.

The flag takes a comma-separated list of key=value pairs mapping 1:1 to the API properties (fullName, username, password, logsInAutomatically, enablesRemoteLogin). It is validated to require a macOS 27+ host and a macOS VM.

The entire user-facing surface is gated behind
'#if arch(arm64) && compiler(>=6.4)' so the flag doesn't appear in help on toolchains that lack the macOS 27 SDK, while the runtime '#available(macOS 27, *)' check gates actual use against the host OS.

### DIFF
--- a/Sources/tart/Commands/Run.swift
+++ b/Sources/tart/Commands/Run.swift
@@ -285,6 +285,28 @@ struct Run: AsyncParsableCommand {
   @Flag(help: ArgumentHelp("Disable the keyboard"))
   var noKeyboard: Bool = false
 
+  #if arch(arm64) && compiler(>=6.4)
+    @Option(help: ArgumentHelp("Provision a macOS guest on first boot using the guest provisioning API", discussion: """
+    Takes a comma-separated list of key=value pairs that configure the initial setup of a macOS guest
+
+    Requires the host to be running macOS 27 (or newer) and only takes effect on the first boot after
+    creation of a macOS 27 (or newer) guest VM.
+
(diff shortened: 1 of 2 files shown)


Sources/tart/Commands/Run.swift
```diff
@@ -285,6 +285,28 @@ struct Run: AsyncParsableCommand {
...
+
+    * fullName=<NAME> — the person's full name to configure
+
+    * username=<USERNAME> — the username for logging into the guest
+
+    * password=<PASSWORD> — the password to configure for the guest
+
+    * logsInAutomatically=true|false — whether to automatically log the person in at startup
+
+    * enablesRemoteLogin=true|false — whether to enable Remote Login (SSH) in the guest
```
torarnv, line 305: I opted to match the `VZMacGuestProvisioningOptions` naming here. We could do `fullname`, or `full_name` or some other scheme if you prefer.

torarnv:
@fkorotkov-oai  :)

motionbug:
Awesome work!

fkorotkov-oai (approved):

fkorotkov-oai:
You are the best!

torarnv:
@fkorotkov-oai Could you cut a new Tart release, that include this change? Thanks!

PR #1265, opened 2026-06-11
Display byte units in text list/get output

### LINKED ISSUE #820: Suggestion: display disk and size attributes with appropriate byte size units
It would be easier to understand disk usage if the disk and VM size exposed by commands like `tart list` and `tart get` were formatted with [appropriate byte modifier](https://developer.apple.com/documentation/foundation/bytecountformatter).

Current:
```
tart list
Source Name           Disk Size State
local  sonoma-vanilla 51   17   stopped
```

Reporting the disk allocation size as 51GB would be slightly easier to grok and match the behavior of similar tools like Docker and Podman:
```
podman system df -v
Images space usage:

REPOSITORY                 TAG         IMAGE ID      CREATED     SIZE        SHARED SIZE  UNIQUE SIZE  CONTAINERS
docker.io/library/alpine   20240329    f8052e6d8143  6 weeks     8.986MB     0B           8.986MB      0
docker.io/library/busybox  latest      23d6e6de0b63  11 months   4.289MB     0B           4.289MB      0

Containers space usage:

CONTAINER ID  IMAGE       COMMAND     LOCAL VOLUMES  SIZE        CREATED     STATUS      NAMES

Local Volumes space usage:

VOLUME NAME                                                       LINKS       SIZE
bfdd925af2aa4ca248f1eb4d2f85a3248ccba91234c5bb5ead4696b50d0e04e4  0           452B
```

For example, the base size of`ghcr.io/cirruslabs/debian:latest` is 1.4GB, but Tart reports it as "1," which is less precise.

It might even be useful to rename Disk to something like "Allocated" since the current terminology can be kind of confusing.

nanookclaw:
Updates the text render paths for `tart list` and `tart get` so their `Disk` and `Size` values are shown with `ByteCountFormatter` human-readable units, while preserving the existing field and column names.

The JSON paths keep their previous machine-readable behavior for compatibility: `tart list --format json` still emits integer GB values for `Disk` and `Size`, and `tart get --format json` still emits integer GB for `Disk` plus the existing three-decimal GB string for `Size`.

Focused tests cover both compatibility shapes and the human-readable text projections.

Fixes #820

## Testing

- `git diff --check`
- Not run: `swift test --filter 'GetTests|ListTests'` (`swift` is not installed in this environment)

### DIFF
--- a/Sources/tart/Commands/Get.swift
+++ b/Sources/tart/Commands/Get.swift
@@ -1,11 +1,76 @@
 import ArgumentParser
 import Foundation
 
-fileprivate struct VMInfo: Encodable {
+struct GetVMInfo: Encodable {
   let OS: OS
   let CPU: Int
   let Memory: UInt64
   let Disk: Int
+  let DiskFormat: String
+  let Size: String
+  let Display: String
+  let Running: Bool
+  let State: String
+
+  private let diskBytes: Int
+  private let sizeBytes: Int
+
+  enum CodingKeys: String, CodingKey {
+    case OS
+    case CPU
+    case Memory
+    case Disk
+    case DiskFormat
+    case Size
+    case Display
+    case Running
+    case State
+  }
+
+  init(
+    OS: OS,
+    CPU: Int,
+    Memory: UInt64,
+    diskBytes: Int,
+    DiskFormat: String,
+    sizeBytes: Int,
+    Display: String,
+    Running: Bool,
+    State: String
+  ) {
+    self.OS = OS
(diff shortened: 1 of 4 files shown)


fkorotkov-oai (approved):

edi-oai (changes requested):
Sorry, perhaps I'm missing the point, but it feels like we can do this much simpler and minimize the overall churn.

What do you think about https://github.com/openai/tart/pull/1301 as an alternative implementation?

PR #1268, opened 2026-06-16
feat(tart): native vmnet networking via VZVmnetNetworkDeviceAttachment

pepicrft:
## What changed

- Added `NetworkVmnet`, a Swift 6.4 gated implementation of Tart's `Network` protocol using vmnet logical networks and `VZVmnetNetworkDeviceAttachment`.
- Added `--net-vmnet` and `--net-vmnet-expose` to `tart run`, including validation for network mode exclusivity and the macOS 26 host requirement reported by the Xcode 27 SDK headers.
- Wired `--net-vmnet-expose` through to `vmnet_network_configuration_add_port_forwarding_rule` for TCP and UDP forwarding.
- When vmnet forwarding is configured, Tart now assigns a stable per-VM IPv4 subnet from the VM MAC, reserves the guest DHCP address, and points forwarding rules at that reserved address.
- Sets `VZVirtualMachineConfiguration.label` on macOS 27+ from the Tart VM name so framework-created system services are easier to identify.
- Added parser coverage for `EXTERNAL_PORT:INTERNAL_PORT[/PROTOCOL]` vmnet forwarding specs and label sanitization coverage for VM names.

## Why

Softnet provides port forwarding through an external sidecar. The newer vmnet logical network APIs let Tart configure shared networking and forwarding in process while still using the standard Virtualization network attachment flow.

While reviewing the WWDC26 Virtualization changes, the VM label API was the other low-risk improvement that fit this PR. It has no user-facing surface area and is safely gated to the Xcode 27/macOS 27 API.

## Impact

Users can opt into native vmnet networking with `--net-vmnet` and can forward host TCP or UDP ports to the guest with `--net-vmnet-expose`. The existing Softnet path remains unchanged.

The vmnet implementation is gated behind `#if compiler(>=6.4)` and `@available(macOS 26, *)`, so the current Swift 6.3/Xcode 26 path still builds while Xcode 27 can type-check the new vmnet APIs. The VM label setter is additionally gated to macOS 27.

## Validation

- `DEVELOPER_DIR=/Applications/Xcode-beta.app/Contents/Developer /usr/bin/xcrun swift build`
- `DEVELOPER_DIR=/Applications/Xcode-beta.app/Contents/Developer /usr/bin/xcrun swift test --filter 'VMTests|NetworkVmnetTests'`
- `/usr/bin/xcrun swift build`
- `/usr/bin/xcrun swift test --filter VMTests`

I have not run a full VM smoke test with `--net-vmnet-expose` yet, so the remaining validation gap is runtime behavior against macOS vmnet.

### DIFF
--- a/Sources/tart/Commands/Run.swift
+++ b/Sources/tart/Commands/Run.swift
@@ -246,11 +246,11 @@ struct Run: AsyncParsableCommand {
   #if compiler(>=6.4)
     @Flag(help: ArgumentHelp("Use native vmnet-backed networking instead of Softnet for port forwarding",
                              discussion: """
-                             Adopts the VZVmnetNetworkDeviceAttachment API introduced in macOS 27 (WWDC26).
+                             Adopts the VZVmnetNetworkDeviceAttachment API introduced in macOS 26.
                              vmnet networks run in-process with no sidecar, so this can replace --net-softnet
                              for the common "CI VM with a few forwarded ports" case.
 
-                             Requires the host to be running macOS 27 (or newer).
+                             Requires the host to be running macOS 26 (or newer).
                              """))
     var netVmnet: Bool = false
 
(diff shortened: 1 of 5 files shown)


fkorotkov-oai:
What is the motivation for this change? Softnet is also using vmnet under the hood and is kind of a superset of what VZVmnetNetworkDeviceAttachment provides. Plus we have much more flexability via a separate process.

PR #1273, opened 2026-06-22
Fix Softnet SUID setup commands

fallintoplace:
## Summary

- probe the exact resolved Softnet executable when checking passwordless sudo
- pass `chown` and `chmod` arguments directly instead of interpolating the executable path into `sh -c`
- restore Tart as the terminal foreground process group after each interactive sudo invocation

## Validation

- `swift build`
- `swift test`

### DIFF
--- a/Sources/tart/Network/Softnet.swift
+++ b/Sources/tart/Network/Softnet.swift
@@ -170,9 +170,9 @@ class Softnet: Network {
       throw SoftnetError.InitializationFailed(why: "\(sudoBinaryName) not found in PATH")
     }
 
-    var process = Process()
+    let process = Process()
     process.executableURL = sudoExecutableURL
-    process.arguments = ["--non-interactive", "softnet", "--help"]
+    process.arguments = ["--non-interactive", softnetExecutablePath, "--help"]
     process.standardInput = nil
     process.standardOutput = nil
     process.standardError = nil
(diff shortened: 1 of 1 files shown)


fkorotkov-oai (dismissed):

edi-oai (changes requested):
Hello Minh 👋 

Could you clarify the threat model this is intended to address?

The implementation appears to treat `PATH` as untrusted while treating paths under the Homebrew Cellar as trusted. On a typical installation, however, the Homebrew prefix is user-owned, so an actor with same-user filesystem access could potentially construct or replace a path that satisfies these checks.

The validation also prevents using a locally built `softnet` through `PATH`.

Would you be open to narrowing the PR to the useful correctness improvements that do not introduce this new trust policy?

In particular, probing the exact resolved `softnet` avoids checking a different binary:

```diff
-process.arguments = ["--non-interactive", "softnet", "--help"]
+process.arguments = ["--non-interactive", softnetExecutablePath, "--help"]
```

The removal of `sh -c` in favor of passing arguments directly through `runInteractiveSudo` also seems worth keeping.

fallintoplace:
I removed the Homebrew/path validation and narrowed the PR to using the resolved Softnet path for the sudo check and running chown/chmod directly instead of through sh -c.
I also fixed the terminal process group handling so each sudo call restores Tart as the foreground process group before returning.
I rebased onto the latest main, and the build and tests pass. Could you take another look when you have time?

edi-oai:
@fallintoplace thanks for the updates, I still think this PR could be simplified quite a bit.

Could you please address the original question in https://github.com/openai/tart/pull/1273#pullrequestreview-4746263346?

fallintoplace:
The threat I had in mind was a manipulated PATH causing Tart to resolve an unexpected Softnet binary, which could then be configured during the sudo setup if the user approved the prompt.

I think that scenario is possible, but treating the user-owned Homebrew prefix as trusted does not address it correctly. I removed that validation and narrowed the PR to using the same resolved Softnet path consistently and avoiding sh -c when running chown and chmod.

Let me know if further adjustments are needed.

PR #1279, opened 2026-07-07
Tart save oci archive

mathstuf:
---
100% vibe coded because I'm not familiar with Swift. I have at least spot-checked the code's logic flow; no idea about idiomatic Swift patterns though.

Tested by exporting an image and copying to DockerHub with `skopeo`'s `oci-archive:` mechanism for the exported file. The blobs copied, but the config failed due to `application/vnd.cirruslabs.tart.config.v1` being an unknown media type.

### DIFF
--- a/Sources/tart/Commands/Save.swift
+++ b/Sources/tart/Commands/Save.swift
@@ -1,0 +1,63 @@
+import ArgumentParser
+import Foundation
+
+struct Save: AsyncParsableCommand {
+  static var configuration = CommandConfiguration(abstract: "Save a VM to an OCI archive file")
+
+  @Argument(help: "local VM name", completion: .custom(completeMachines))
+  var localName: String
+
+  @Argument(help: "output archive path", completion: .file())
+  var path: String
+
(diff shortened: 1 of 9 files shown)


mathstuf:
This at least gets to DockerHub. Need to test that pulling the image from there works.

mathstuf:
Pulling the image fails with:

```
tart command returned non-zero exit code: "Error: DecodingError.dataCorrupted: Data was corrupted. Debug description: The given data was not valid JSON.. Underlying error: Error Domain=NSCocoaErrorDomain Code=3840 \"Unexpected character '<' around line 1, column 1.\" UserInfo={NSJSONSerializationErrorIndex=0, NSDebugDescription=Unexpected character '<' around line 1, column 1.}"
```

This happens with both `skopeo copy` between a GitLab registry and DockerHub and the oci-archive export of this PR and DockerHub. I'm partial to blaming `skopeo` here. I plan on using `dive` to investigate the two, but it'll take time to pull the data down.

PR #1284, opened 2026-07-17
Fix Tart release code signing

fkorotkov-oai:
## Summary

Both attempts of the Tart 2.33.0 release built successfully, then spent roughly 80 minutes in `.ci/sign-release.sh` until the job timed out; cleanup reported an orphaned `codesign` process. The hosted `macos-26-arm64` runner also warns that the installed `gon` binary does not support ARM64.

This removes the archived `gon` dependency and signs the existing app bundle directly with `codesign`, explicitly selecting the temporary keychain and preserving the hardened-runtime and entitlement settings. It also keeps the keychain unlocked for the duration of the release, registers it in the user search list, verifies the resulting signature, and exposes signing-hook output in GoReleaser. The obsolete `gon.hcl` and unused `AC_PASSWORD` exposure are removed.

## Validation

- `git diff --check`
- `sh -n .ci/sign-release.sh`
- parsed `.goreleaser.yml` and `.github/workflows/release.yml` with `Psych.parse_file`
- exercised both release and snapshot signing branches with a stub `codesign`
- validated the GoReleaser schema with the pre-existing deprecated `brews` section isolated

The existing `brews` deprecation is unchanged.


edi-oai (approved):

vsncls:
2.33 supplied app bundle is not notarized : 

https://github.com/openai/homebrew-tools/issues/27#issue-4914642408

PR #1287, opened 2026-07-21
Pass Softnet policy control FD through Tart

fkorotkov-oai:
## Summary

Passes a dedicated Softnet policy-control socket through `tart run`, enabling callers such as Orchard to dynamically update allow/block lists while leaving the existing VM packet socket unchanged. The forwarded protocol is intentionally limited to `softnet.policy.get` and `softnet.policy.set`.

Depends on https://github.com/openai/softnet/pull/181.

## Changes

- add `--net-softnet-control-fd` and imply Softnet networking when appropriate (including host-network support)
- require a connected Unix stream socket above the standard descriptors
- map the inherited descriptor to the Softnet child's stdout and pass `--control-fd 1`; stdin remains the existing Unix datagram packet socket
- close the parent-side descriptor on validation, initialization, launch, and normal cleanup paths
- add an executable fake-Softnet test that verifies exact arguments, socket types, descriptor handoff, and duplex exchange using the `softnet.policy.set` terminology

## Validation

- 10 focused `SoftnetControlFDTests` passed in a clean export
- SwiftFormat lint passed
- cross-repo stale-RPC-reference check and `git diff --check`

The clean export intentionally excludes unrelated, pre-existing local Windows/QEMU files in the checkout.

### DIFF
--- a/Sources/tart/Commands/Run.swift
+++ b/Sources/tart/Commands/Run.swift
@@ -224,6 +224,13 @@ struct Run: AsyncParsableCommand {
   """, valueName: "comma-separated CIDRs"))
   var netSoftnetBlock: String?
 
+  @Option(help: ArgumentHelp("Connected Unix stream socket file descriptor to use for the Softnet control channel (e.g. --net-softnet-control-fd=3)", discussion: """
+  This option enables the Softnet control channel on an inherited Unix stream socket. It can be used to dynamically replace Softnet allow and block lists while the VM is running.
+
+  The file descriptor must be greater than 2. Implies --net-softnet unless --net-host is specified.
+  """, valueName: "file descriptor"))
+  var netSoftnetControlFd: Int32?
+
   @Option(help: ArgumentHelp("Comma-separated list of TCP ports to expose (e.g. --net-softnet-expose 2222:22,8080:80)", discussion: """
   Options are comma-separated and are as follows:
 
@@ -313,7 +320,7 @@ struct Run: AsyncParsableCommand {
     }
 
     // Automatically enable --net-softnet when any of its related options are specified
-    if netSoftnetAllow != nil || netSoftnetBlock != nil || netSoftnetExpose != nil {
+    if netSoftnetAllow != nil || netSoftnetBlock != nil || netSoftnetExpose != nil || (netSoftnetControlFd != nil && !netHost) {
       netSoftnet = true
     }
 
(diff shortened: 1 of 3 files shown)


Sources/tart/Commands/Run.swift
```diff
@@ -313,7 +320,7 @@ struct Run: AsyncParsableCommand {
     }
 
     // Automatically enable --net-softnet when any of its related options are specified
-    if netSoftnetAllow != nil || netSoftnetBlock != nil || netSoftnetExpose != nil {
+    if netSoftnetAllow != nil || netSoftnetBlock != nil || netSoftnetExpose != nil || (netSoftnetControlFd != nil && !netHost) {
```
edi-oai: Why special-case `netSoftnetControlFd`?

Like the other Softnet-related options, it should enable `netSoftnet`, allowing the existing validation to reject its use with `netHost`:

```patch
diff --git a/Sources/tart/Commands/Run.swift b/Sources/tart/Commands/Run.swift
@@
-  The file descriptor must be greater than 2. Implies --net-softnet unless --net-host is specified.
+  The file descriptor must be greater than 2. Implies --net-softnet.
@@
-    if netSoftnetAllow != nil || netSoftnetBlock != nil || netSoftnetExpose != nil || (netSoftnetControlFd != nil && !netHost) {
+    if netSoftnetAllow != nil || netSoftnetBlock != nil || netSoftnetExpose != nil || netSoftnetControlFd != nil {
       netSoftnet = true
     }

diff --git a/Tests/TartTests/SoftnetControlFDTests.swift b/Tests/TartTests/SoftnetControlFDTests.swift
@@
-  func testControlFDWorksWithHostNetworking() throws {
+  func testControlFDIsRejectedWithHostNetworking() throws {
     let temporaryHome = try createTemporaryTartHome()
     defer { try? FileManager.default.removeItem(at: temporaryHome) }
     let previousHome = ProcessInfo.processInfo.environment["TART_HOME"]
     setenv("TART_HOME", temporaryHome.path, 1)
     defer { restoreEnvironment("TART_HOME", value: previousHome) }

-    let command = try Run.parse(["vm", "--net-host", "--net-softnet-control-fd", "3"])
-
-    XCTAssertTrue(command.netHost)
-    XCTAssertFalse(command.netSoftnet)
-    XCTAssertEqual(command.netSoftnetControlFd, 3)
+    XCTAssertThrowsError(
+      try Run.parse(["vm", "--net-host", "--net-softnet-control-fd", "3"])
+    )
   }
```

Sources/tart/Network/Softnet.swift
```diff
@@ -13,10 +13,22 @@ class Softnet: Network {
   private let process = Process()
   private var monitorTask: Task<Void, Error>? = nil
   private let monitorTaskFinished = ManagedAtomic<Bool>(false)
+  private var controlFD: Int32?
 
   let vmFD: Int32
 
-  init(vmMACAddress: String, extraArguments: [String] = []) throws {
+  init(vmMACAddress: String, extraArguments: [String] = [], controlFD: Int32? = nil) throws {
+    if let controlFD = controlFD {
```
edi-oai, line 22: Could we use an owning `FileHandle` with `closeOnDealloc: true` here?

That would remove the manual `closeControlFD()`/`deinit` bookkeeping while still allowing `run()` to close the parent’s copy after launching Softnet.

Sources/tart/Commands/Run.swift
fkorotkov-oai: Addressed in 311eec9. `--net-softnet-control-fd` now consistently implies `--net-softnet`, so combining it with `--net-host` is rejected by the existing mutual-exclusion validation. Updated the help text and test accordingly. Full Swift test suite passes (56 tests).

Sources/tart/Network/Softnet.swift
fkorotkov-oai: Addressed in 4a018ce. The control socket now uses an owning `FileHandle(closeOnDealloc: true)`; `run()` explicitly closes the parent copy after launch, and the raw-FD/deinit bookkeeping is gone. Standard descriptors are rejected before ownership is taken, with regression coverage to ensure stderr remains open. Full Swift test suite and SwiftFormat lint pass.

edi-oai (approved):

PR #1296, opened 2026-07-28
Add optional @PPI hint to --display for HiDPI pixel displays

jonnyzzz:
## Problem

`tart set --display WIDTHxHEIGHTpx` builds the guest's pixel-unit framebuffer with a **hardcoded 72 PPI** (`Sources/tart/Platform/Darwin.swift`). As a result a macOS guest never exposes a HiDPI / Retina (`scaling:on`, 2x) mode when the host display is non-Retina — or when there is no host display at all, which is the normal case on headless / CI hosts. The point-unit (`pt`) path can produce a HiDPI guest, but only by copying the host `NSScreen.main` backing scale, so it depends on a Retina host console being attached and a GUI session being logged in. There was previously no way to ask for a Retina guest on such a host.

## Change

Allow an optional `@PPI` (pixels-per-inch) suffix on `--display`:

```
tart set my-vm --display 3200x1800px@220
```

At a Retina-class density (e.g. `220`) the pixel display then offers a `scaling:on` (HiDPI, 2x) mode inside the guest. Because the pixel path never consults `NSScreen`, this works **regardless of the host display or whether a user is logged in on the host** — it only needs the (already-required) unlocked login keychain that Virtualization.framework needs to boot any VM on macOS 15+.

- New optional field `VMDisplayConfig.ppi` (persisted in `config.json`; omitted when nil, so existing configs are untouched).
- `--display` parser accepts `WIDTHxHEIGHT[pt|px][@PPI]`; the `@PPI` is parsed before the unit suffix and degrades to “unset” on malformed input.
- `VMDisplayConfig.effectivePixelsPerInch` returns `ppi ?? 72`, so **behavior is unchanged when the hint is omitted**.
- `CustomStringConvertible` round-trips the hint (`3200x1800px@220`).

The hint applies to the pixel path; on the point path it is stored but not used (that path derives scale from the host screen).

## Tests

Added to `Tests/TartTests/VMConfigTests.swift` (TDD — each was watched failing before implementation):

- parses `@PPI` suffix (with and without an explicit unit);
- description round-trips the hint;
- `effectivePixelsPerInch` uses the configured value and defaults to 72 when unset;
- backward compatibility: existing `WIDTHxHEIGHT[pt|px]` strings carry no PPI;
- malformed PPI (`@notanumber`) degrades to the 72 default rather than failing the parse.

Full suite green (`swift test`), and `swiftformat --lint` reports the changed files clean.

### DIFF
--- a/Sources/tart/Commands/Set.swift
+++ b/Sources/tart/Commands/Set.swift
@@ -14,7 +14,7 @@ struct Set: AsyncParsableCommand {
   @Option(help: "VM memory size in megabytes")
   var memory: UInt64?
 
-  @Option(help: "VM display resolution in a format of WIDTHxHEIGHT[pt|px]. For example, 1200x800, 1200x800pt or 1920x1080px. Units are treated as hints and default to \"pt\" (points) for macOS VMs and \"px\" (pixels) for Linux VMs when not specified.")
+  @Option(help: "VM display resolution in a format of WIDTHxHEIGHT[pt|px][@PPI]. For example, 1200x800, 1200x800pt, 1920x1080px or 3200x1800px@220. Units are treated as hints and default to \"pt\" (points) for macOS VMs and \"px\" (pixels) for Linux VMs when not specified. The optional @PPI (pixels-per-inch) hint applies to pixel-unit displays and, at a Retina-class value such as 220, makes a macOS guest expose a HiDPI (2x) mode regardless of the host display; it defaults to 72 (non-Retina) when omitted.")
   var display: VMDisplayConfig?
 
   @Flag(inversion: .prefixedNo, help: ArgumentHelp("Whether to automatically reconfigure the VM's display to fit the window"))
@@ -57,6 +57,7 @@ struct Set: AsyncParsableCommand {
         vmConfig.display.height = display.height
       }
       vmConfig.display.unit = display.unit
+      vmConfig.display.ppi = display.ppi
     }
 
     vmConfig.displayRefit = displayRefit
@@ -91,6 +92,14 @@ extension VMDisplayConfig: ExpressibleByArgument {
   public init(argument: String) {
     var argument = argument
     var unit: Unit? = nil
+    var ppi: Int? = nil
+
+    // Optional "@PPI" pixels-per-inch hint, e.g. "3200x1800px@220". Parsed
+    // before the unit suffix since it always trails the whole spec.
+    if let atIndex = argument.lastIndex(of: "@") {
+      ppi = Int(argument[argument.index(after: atIndex)...])
+      argument = String(argument[..<atIndex])
+    }
 
     if argument.hasSuffix(Unit.pixel.rawValue) {
       argument = String(argument.dropLast(Unit.pixel.rawValue.count))
@@ -107,6 +116,7 @@ extension VMDisplayConfig: ExpressibleByArgument {
       width: parts[safe: 0] ?? 0,
       height: parts[safe: 1] ?? 0,
       unit: unit,
+      ppi: ppi,
     )
   }
 }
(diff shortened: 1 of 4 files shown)


jonnyzzz:
### Validated end-to-end on a real macOS guest

Booted `ghcr.io/cirruslabs/macos-sequoia-base:latest` (macOS 15.7.7) twice with a locally-built, ad-hoc-signed `tart` from this branch and read the guest's CoreGraphics main-display mode over SSH:

| `--display` | guest current mode (points) | backing (pixels) | HiDPI |
|---|---|---|---|
| `3200x1800px` (unchanged, 72 PPI) | 3200×1800 | 3200×1800 | no (1×) |
| `3200x1800px@220` | **1600×900** | **3200×1800** | **yes (2×)** |

So the `@220` hint makes the guest boot straight into a 2×/Retina mode (1600×900 logical, 3200×1800 backing), whereas the current behavior offers no HiDPI mode at all. The pixel path never touches `NSScreen`, so this is independent of the host display — the same VM went Retina on an ordinary host with no Retina console attached and no reliance on a logged-in GUI session.

Sources/tart/Commands/Set.swift
```diff
@@ -57,6 +57,7 @@ struct Set: AsyncParsableCommand {
         vmConfig.display.height = display.height
       }
       vmConfig.display.unit = display.unit
+      vmConfig.display.ppi = display.ppi
```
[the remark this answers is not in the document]
jonnyzzz: Good catch — fixed in 60bf256. The `--display` parser now keeps only a positive PPI (`@0`/`@-1` are ignored like non-numeric input), and `effectivePixelsPerInch` also treats a non-positive stored value (e.g. a hand-edited config) as unset, so a non-positive value can never reach `VZMacGraphicsDisplayConfiguration` — it falls back to 72. Covered by two new tests in VMConfigTests.

jonnyzzz:
@codex review

Sources/tart/Platform/Darwin.swift
```diff
@@ -95,9 +95,11 @@ struct UnsupportedHostOSError: Error, CustomStringConvertible {
         VZMacGraphicsDisplayConfiguration(
           widthInPixels: vmConfig.display.width,
           heightInPixels: vmConfig.display.height,
-          // A reasonable guess according to Apple's documentation[1]
+          // Defaults to 72 — a reasonable guess according to Apple's
+          // documentation[1] — unless a --display "@PPI" hint asks for a
+          // Retina-class density (which yields a HiDPI guest mode).
           // [1]: https://developer.apple.com/documentation/coregraphics/1456599-cgdisplayscreensize
-          pixelsPerInch: 72
+          pixelsPerInch: vmConfig.display.effectivePixelsPerInch
```
fkorotkov-oai, line 102: The `NSScreen.main` thing above should configure DPI correctly. There is some history around this issue.

fkorotkov-oai: Some history behind the 72-PPI default:

- #22 introduced manual display/DPI settings (defaulting to 72) after Retina display sizing produced impractically small UI.
- #106 fixed blurry rendering reported in #104 by replacing manual DPI with host-screen inference via `NSScreen.main`. It retained 72 only as the no-screen fallback, citing [Apple's convention of assuming 72 DPI when EDID information is unavailable](https://developer.apple.com/documentation/coregraphics/cgdisplayscreensize%28_%3A%29).
- #1155 later added `pt`/`px`: points retained host-derived density, while explicit pixels reused that existing 72-PPI fallback.

So host DPI inheritance already covers the point-unit path when a screen is available. This PR adds an explicit override for the pixel/headless path; keeping 72 as its default preserves existing behavior.

Issue #1269, opened 2026-06-16
Design layered ASIF images for CI image variants


pepicrft:
## Context

Tart's current model is intentionally simple:

- A local VM is a directory with `config.json`, `nvram.bin`, and `disk.img`.
- A remote VM is an OCI reference, for example `ghcr.io/org/image:tag`.
- `tart clone` creates a local VM and relies on APFS copy-on-write when available.
- Tart's current OCI disk layers are transfer chunks, not semantic VM image layers like "base macOS", "Xcode", or "iOS simulator runtime".

That works well for local clones, but CI image distribution has a different pressure point. Images such as "macOS + Xcode" and "macOS + Xcode + simulator runtime" share most of their disk content, but today we still tend to publish and consume them as full image variants.

WWDC26 introduced DiskImageKit stacked ASIF images and `VZDiskImageStorageDeviceAttachment(diskImage:)`. That gives Tart a public Apple API for block-level image stacks:

```text
base ASIF image
+ read-only child ASIF image
+ writable overlay ASIF image
```

This is close to the image/template layering model used by systems such as Anka, VMware linked clones, Parallels linked clones, Hyper-V differencing disks, QEMU backing chains, and Docker/OCI image layers. The common pattern is that users think in images, templates, or tags, while layers are the storage and distribution mechanism.

## Proposal

Keep Tart's user-facing concepts as **VMs** and **images**. Do not introduce `tart layer` as a top-level concept initially.

Use **layers** internally as immutable ASIF artifacts referenced by image manifests and cached by digest.

A CI image variant would look like this internally:

```text
A: macOS 27 + Xcode 27
B: delta from A after installing iOS 27 simulator runtime
C: delta from A after installing watchOS 27 simulator runtime
D: delta from B after installing watchOS 27 simulator runtime
```

Registry tags would point at ordered chains:

```text
ghcr.io/tuist/macos-xcode:27              -> A
ghcr.io/tuist/macos-xcode:27-ios-27       -> A + B
ghcr.io/tuist/macos-xcode:27-watch-27     -> A + C
ghcr.io/tuist/macos-xcode:27-ios-watch-27 -> A + B + D
```

Consumers pull image tags, not individual layers:

```bash
tart pull ghcr.io/tuist/macos-xcode:27-ios-27
tart run --ephemeral ghcr.io/tuist/macos-xcode:27-ios-27
```

[...]

PR #1303, opened 2026-08-05
Begin work to support DiskImageKit in tart

yzhuang-oai:
This is first of a couple of PRs to add support for the new
DiskImageKit ASIF layers to tart VM images.

https://developer.apple.com/documentation/diskimagekit

This change is focused on laying down the OCI media type for
ASIF layers, the content addressable store structure,
DiskImageStack type, as well as the VMDirectory structure
for supporting layers.

Since this is a larger feature I am doing this in smaller PRs to help
with review.

Issue: #1269

### DIFF
--- a/Sources/tart/ContentStore.swift
+++ b/Sources/tart/ContentStore.swift
@@ -1,0 +1,94 @@
+import Foundation
+
+enum ContentStoreError: Error, Equatable {
+  case invalidContentDigest(String)
+  case contentDigestMismatch(expected: String, actual: String)
+}
+
+/// Opaque content-addressed storage for immutable reconstructed files.
+///
+/// Stacked disks currently use it for complete base disks and published ASIF
+/// overlays reconstructed from Tart disk chunks. OCI blob digests may differ
+/// across registries, so the key is the full reconstructed-file digest.
+struct ContentStore {
+  private static let digestAlgorithm = "sha256"
+  private static let digestPrefix = "\(digestAlgorithm):"
+
+  let baseURL: URL
+  private let digestDirectoryURL: URL
+
+  init() throws {
+    try self.init(baseURL: Config().tartCacheDir.appendingPathComponent("content", isDirectory: true))
+  }
+
+  init(baseURL: URL) throws {
+    self.baseURL = baseURL
+    self.digestDirectoryURL = baseURL.appendingPathComponent(Self.digestAlgorithm, isDirectory: true)
+    try FileManager.default.createDirectory(at: digestDirectoryURL, withIntermediateDirectories: true)
+  }
+
+  func contentURL(for contentDigest: String) throws -> URL {
+    let digestHex = try validatedDigestHex(contentDigest)
+
+    return digestDirectoryURL.appendingPathComponent(digestHex)
+  }
+
+  func temporaryContentURL(for contentDigest: String) throws -> URL {
+    let targetURL = try contentURL(for: contentDigest)
+
+    return targetURL.deletingLastPathComponent().appendingPathComponent(".\(UUID().uuidString).tmp")
+  }
+
+  /// Returns a validated cache hit. Corrupt files are treated as misses so a
+  /// later pull can safely rebuild them.
+  func existingContentURL(for contentDigest: String) throws -> URL? {
+    let url = try contentURL(for: contentDigest)
+
+    guard FileManager.default.fileExists(atPath: url.path) else {
+      return nil
+    }
+
+    guard try Digest.hash(url) == contentDigest else {
+      return nil
+    }
+
+    return url
+  }
+
+  /// Move a fully reconstructed temporary file into the cache after verifying
+  /// its semantic identity. The caller should create the temporary file with
+  /// `temporaryContentURL(for:)` so rename stays on the same filesystem.
+  func install(_ temporaryURL: URL, contentDigest: String) throws -> URL {
+    let actualDigest = try Digest.hash(temporaryURL)
+    guard actualDigest == contentDigest else {
+      throw ContentStoreError.contentDigestMismatch(expected: contentDigest, actual: actualDigest)
+    }
+
+    let targetURL = try contentURL(for: contentDigest)
+
+    if let existingURL = try existingContentURL(for: contentDigest) {
+      try? FileManager.default.removeItem(at: temporaryURL)
+      return existingURL
+    }
+
+    try? FileManager.default.removeItem(at: targetURL)
+    try FileManager.default.moveItem(at: temporaryURL, to: targetURL)
+
+    return targetURL
+  }
+
+  private func validatedDigestHex(_ contentDigest: String) throws -> String {
+    guard contentDigest.hasPrefix(Self.digestPrefix) else {
+      throw ContentStoreError.invalidContentDigest(contentDigest)
+    }
+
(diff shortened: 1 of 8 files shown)


Sources/tart/DiskLayerCache.swift
```diff
@@ -0,0 +1,137 @@
...
+/// chunks or an ASIF overlay reconstructed from overlay chunks. OCI blob
+/// digests may differ across registries, so the cache key is the full
+/// reconstructed-file digest.
+struct DiskLayerCache {
+  private static let digestPrefix = "sha256:"
+
+  let baseURL: URL
+
+  init() throws {
+    try self.init(baseURL: Config().tartCacheDir.appendingPathComponent("OCIs/layers", isDirectory: true))
```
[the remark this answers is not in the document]
yzhuang-oai: This will be addressed once `DiskLayerCache` is integrated into the CLI commands, right now it is just creating this type and we don't need to worry about pruning at the moment.

Sources/tart/DiskLayerCache.swift
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    return url
+  }
+
+  /// Move a fully reconstructed temporary file into the cache after verifying
+  /// its semantic identity. The caller should create the temporary file with
+  /// `temporaryLayerURL(for:)` so rename stays on the same filesystem.
+  func install(_ temporaryURL: URL, contentDigest: String) throws -> URL {
+    let actualDigest = try Digest.hash(temporaryURL)
```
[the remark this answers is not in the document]
yzhuang-oai: Fixed.

Sources/tart/OCI/Manifest.swift
```diff
@@ -7,18 +7,48 @@ let ociConfigMediaType = "application/vnd.oci.image.config.v1+json"
...
 let configMediaType = "application/vnd.cirruslabs.tart.config.v1"
 let diskV2MediaType = "application/vnd.cirruslabs.tart.disk.v2"
+let asifOverlayMediaType = "application/vnd.cirruslabs.tart.disk.asif.overlay.v1"
 let nvramMediaType = "application/vnd.cirruslabs.tart.nvram.v1"
 
 // Manifest annotations
 let uncompressedDiskSizeAnnotation = "org.cirruslabs.tart.uncompressed-disk-size"
 let uploadTimeAnnotation = "org.cirruslabs.tart.upload-time"
+let diskBlockSizeAnnotation = "org.cirruslabs.tart.disk.block-size"
+let diskBlockCountAnnotation = "org.cirruslabs.tart.disk.block-count"
```
edi-oai: Do you think we can dynamically derive the block count from `uncompressed-disk-size / block-size`?

This could slightly help to reduce manifest's surface.

Sources/tart/ContentStore.swift
```diff
@@ -0,0 +1,95 @@
...
+    let digestHex = try validatedDigestHex(contentDigest)
+
+    return baseURL
+      .appendingPathComponent("sha256", isDirectory: true)
+      .appendingPathComponent(digestHex)
+  }
+
+  func temporaryContentURL(for contentDigest: String) throws -> URL {
+    let targetURL = try contentURL(for: contentDigest)
+    try FileManager.default.createDirectory(at: targetURL.deletingLastPathComponent(), withIntermediateDirectories: true)
```
edi-oai: Can we create this `sha256` directory just once in `init()`?

I also see the same pattern [on line `68`](https://github.com/openai/tart/pull/1303/changes#diff-42eb4ab67ad72a0e051524acb5e9db769364625e1e7c75d2a1565ba92f5f9100R68).

Sources/tart/VMDirectory.swift
```diff
@@ -87,10 +90,47 @@ struct VMDirectory: Prunable {
...
+  private var hasRequiredMetadata: Bool {
+    let fileManager = FileManager.default
+
+    return fileManager.fileExists(atPath: configURL.path) &&
+      fileManager.fileExists(atPath: nvramURL.path)
+  }
+
+  /// Existing Tart layout with one independently attachable `disk.img`.
+  /// A pulled flat OCI record may also carry `manifest.json`.
+  var isStandalone: Bool {
```
edi-oai: What do you think about representing this, `isStackedLocal` and `isStackedOCIRecord` as an `enum`?

For example:

```swift
enum Layout: Equatable {
  case standalone
  case stackedLocal
  case stackedOCIRecord

  var isRunnable: Bool {
    self != .stackedOCIRecord
  }
}
```

One benefit is that we don't have to repeat file checks when switching on it.

Sources/tart/DiskImageStack.swift
```diff
@@ -0,0 +1,290 @@
...
+      }
+
+      let image = try DiskImage(opening: .open(url: url, mode: mode))
+      guard image.format == .asif else {
+        throw DiskImageStackError.invalidDiskImage(url, "overlay must use ASIF format")
+      }
+      guard let overlayUUID = image.layerUUID else {
+        throw DiskImageStackError.invalidDiskImage(url, "ASIF overlay is missing a UUID")
+      }
+      if let previousUUID, image.parentUUID != previousUUID {
```
edi-oai: What do you think about relying on `DiskImageKit` to validate stack compatibility when appending an overlay?

For example:

```swift
do {
  return try parent.appending(overlay)
} catch is IncompatibleStackingError {
  throw DiskImageStackError.invalidDiskImage(
    url,
    "ASIF overlay is incompatible with its parent"
  )
}
```

One benefit is that we don't have to manually carry and compare the parent UUID, block size, and block count.

This could also let `openOverlay` return a `DiskImage` directly, and `validatedParentImage` return only the assembled image instead of carrying UUID and block-count tuples.

Tart can keep validating its own metadata, such as the content digest, overlay role, and final manifest geometry.

edi-oai (dismissed):
LGTM overall, left a couple of nits and questions.

Sources/tart/OCI/Manifest.swift
yzhuang-oai: Thanks, fixed.

Sources/tart/ContentStore.swift
yzhuang-oai: Thanks, fixed.

Sources/tart/VMDirectory.swift
yzhuang-oai: Sounds good, fixed.

Sources/tart/DiskImageStack.swift
yzhuang-oai: Done.

Sources/tart/ContentStore.swift
```diff
@@ -0,0 +1,94 @@
...
+  /// Returns a validated cache hit. Corrupt files are treated as misses so a
+  /// later pull can safely rebuild them.
+  func existingContentURL(for contentDigest: String) throws -> URL? {
+    let url = try contentURL(for: contentDigest)
+
+    guard FileManager.default.fileExists(atPath: url.path) else {
+      return nil
+    }
+
+    guard try Digest.hash(url) == contentDigest else {
```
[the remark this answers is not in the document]
yzhuang-oai: This will be addressed in a later change.

edi-oai (approved):

PR #1304, opened 2026-08-06
Add OCI transport and base clone with DiskImageKit 

yzhuang-oai:
This is the second PR adding support for new
DiskImageKit ASIF layers to tart VM images.

https://developer.apple.com/documentation/diskimagekit

This change implements the OCI transport logic and 
adds the `tart clone --stacked` option to clone an existing
VM image into the new stacked VM image format.

Issue: #1269


Sources/tart/VMDirectory.swift
```diff
@@ -130,10 +130,13 @@ struct VMDirectory: Prunable {
   }
 
   var initialized: Bool {
-    // Layered layouts are recognized above so pull/clone code can distinguish
-    // them, but they are not runnable until the VM lifecycle paths consume
-    // manifest.json and overlay.asif through DiskImageStack.
-    isStandalone
+    isStandalone || isLayeredLocal
+  }
```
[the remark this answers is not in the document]
yzhuang-oai: Fixed.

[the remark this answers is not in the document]
yzhuang-oai: Fixed.

Sources/tart/VMStorageOCI.swift
```diff
@@ -141,7 +185,9 @@ class VMStorageOCI: PrunableStorage {
   }
 
   func prunables() throws -> [Prunable] {
-    try list().filter { (_, _, isSymlink) in !isSymlink }.map { (_, vmDir, _) in vmDir }
+    try list().filter { (_, vmDir, isSymlink) in
+      !isSymlink && vmDir.isStandalone
+    }.map { (_, vmDir, _) in vmDir }
```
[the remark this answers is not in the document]
yzhuang-oai: Pruning will be done in a layer PR addressing stale layer content.

Sources/tart/ContentStore.swift
```diff
@@ -39,15 +41,50 @@ struct ContentStore {
...
   }
 
-  /// Returns a validated cache hit. Corrupt files are treated as misses so a
-  /// later pull can safely rebuild them.
-  func existingContentURL(for contentDigest: String) throws -> URL? {
+  /// Returns a stable staging path so an interrupted registry pull can resume
+  /// reconstructing this content entry on a later attempt.
+  func resumableContentURL(for contentDigest: String) throws -> URL {
+    let targetURL = try contentURL(for: contentDigest)
+    try FileManager.default.createDirectory(at: targetURL.deletingLastPathComponent(), withIntermediateDirectories: true)
```
edi-oai: Is this redundant?

I think `ContentStore.init(baseURL:)` already pre-creates `digestDirectoryURL`.

Sources/tart/ContentStore.swift
```diff
@@ -39,15 +41,50 @@ struct ContentStore {
...
+
+    return targetURL.deletingLastPathComponent().appendingPathComponent(".\(targetURL.lastPathComponent).partial")
+  }
+
+  /// Returns a stable lock file for serializing reconstruction of one content
+  /// entry. The file is intentionally retained; flock state lives on the file
+  /// descriptor and disappears when the owning process exits.
+  func lockURL(for contentDigest: String) throws -> URL {
+    let targetURL = try contentURL(for: contentDigest)
+    try FileManager.default.createDirectory(at: targetURL.deletingLastPathComponent(), withIntermediateDirectories: true)
```
edi-oai: Is this redundant?

I think `ContentStore.init(baseURL:)` already pre-creates `digestDirectoryURL`.

Sources/tart/Commands/Clone.swift
```diff
@@ -31,6 +31,9 @@ struct Clone: AsyncParsableCommand {
   @Flag(help: .hidden)
   var deduplicate: Bool = false
 
+  @Flag(help: "create a stacked disk that uses the source image as an immutable base")
+  var base: Bool = false
```
fkorotkov-oai: What do you and @edi-oai think about using `stacked` instead? Kind of exposing that. `--base` is related to the subject of cloning but `--stacked` will indicate difference in the result of the operation.

fkorotkov-oai (commented):
Did only a quick check and seems there are two issues to double check:

* interrupted pull will leave content
* pull of a flat image before a pull of a stacked image will not reuse flat `disk.img`.

Sources/tart/ContentStore.swift
```diff
@@ -66,15 +104,54 @@ struct ContentStore {
 
     let targetURL = try contentURL(for: contentDigest)
 
-    if let existingURL = try existingContentURL(for: contentDigest) {
-      try? FileManager.default.removeItem(at: temporaryURL)
-      return existingURL
+    while true {
```
edi-oai: Can we just acquire a content-store-wide directory lock inside `install()` to simplify this?

For example:

```swift
let lock = try FileLock(lockURL: baseURL)
try lock.lock()
defer { try? lock.unlock() }
```

Then just proceed assuming that no other `install()` will race with us.

Sources/tart/ContentStore.swift
```diff
@@ -66,15 +104,54 @@ struct ContentStore {
...
+      if try exchangeItem(at: temporaryURL, with: targetURL) {
+        try? FileManager.default.removeItem(at: temporaryURL)
+        return targetURL
+      }
+    }
+  }
+
+  /// Atomically publishes a content entry without replacing an existing one.
+  /// Returns false when another installer already created the destination.
+  private func moveItemWithoutReplacing(at sourceURL: URL, to destinationURL: URL) throws -> Bool {
```
edi-oai: We can remove this function if we do https://github.com/openai/tart/pull/1304/changes#r3761967194.

Sources/tart/ContentStore.swift
```diff
@@ -66,15 +104,54 @@ struct ContentStore {
...
+      return false
+    }
+
+    let details = Errno(rawValue: CInt(errno))
+    throw ContentStoreError.operationFailed("failed to install content entry \(destinationURL.path): \(details)")
+  }
+
+  /// Atomically exchanges a verified temporary file with a corrupt content
+  /// entry. Returns false when the destination disappeared before the swap.
+  private func exchangeItem(at sourceURL: URL, with destinationURL: URL) throws -> Bool {
```
edi-oai: We can remove this function if we do https://github.com/openai/tart/pull/1304/changes#r3761967194.

Sources/tart/DiskAttachmentSource.swift
```diff
@@ -0,0 +1,28 @@
+import Foundation
```
edi-oai: Can we just pass `vmDir` to `craftConfiguration()`, and then just branch there on `vmDir.isStackedVM`?

This would allow us to not introduce this file at all.

Sources/tart/Commands/Run.swift
```diff
@@ -952,14 +957,27 @@ struct VMView: NSViewRepresentable {
...
+    )
   }
 
-  static func craft(_ diskPath: String, readOnly diskReadOnly: Bool, syncModeRaw: String, cachingModeRaw: String) throws -> VZStorageDeviceConfiguration {
+  private static func craft(
+    _ diskPath: String,
+    readOnly diskReadOnly: Bool,
+    syncModeRaw: String,
+    cachingModeRaw: String
+  ) throws -> (VZStorageDeviceConfiguration, FileLock?) {
```
edi-oai: Can we just return `AdditionalDisk` here?

By having a private constructor in `AdditionalDisk`:

```swift
private init(
  configuration: VZStorageDeviceConfiguration,
  temporaryDiskLock: FileLock? = nil
) {
  self.configuration = configuration
  self.temporaryDiskLock = temporaryDiskLock
}
```

Sources/tart/VMStorageOCI.swift
```diff
@@ -181,11 +319,13 @@ class VMStorageOCI: PrunableStorage {
       throw CancellationError()
     }
 
-    if !exists(digestName) {
+    if try !hasCompleteCachedImage(digestName, manifest: manifest) {
```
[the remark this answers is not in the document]
yzhuang-oai: fixed in: https://github.com/openai/tart/pull/1304/changes/b78f26797ae4caffbf2f9957dfcc0020935a1009

Sources/tart/Commands/Clone.swift
yzhuang-oai: `--stacked` works for me, fixed in https://github.com/openai/tart/pull/1304/changes/b78f26797ae4caffbf2f9957dfcc0020935a1009

Sources/tart/ContentStore.swift
yzhuang-oai: Fixed.

Sources/tart/ContentStore.swift
yzhuang-oai: Thanks, fixed.

Sources/tart/ContentStore.swift
yzhuang-oai: Thanks, fixed.

Sources/tart/ContentStore.swift
yzhuang-oai: Thanks, fixed.

Sources/tart/ContentStore.swift
yzhuang-oai: Fixed.

Sources/tart/Commands/Run.swift
yzhuang-oai: Thanks, fixed.

Sources/tart/DiskAttachmentSource.swift
yzhuang-oai: Fixed.

yzhuang-oai:
> Did only a quick check and seems there are two issues to double check:
> 
> * interrupted pull will leave content
> * pull of a flat image before a pull of a stacked image will not reuse flat `disk.img`.

For interrupted pull will leave content, this is intentional to leave a digest-scoped .partial file so a later pull can resume instead of redownloading every OCI chunk. This GC/content store prune will ensure it gets cleaned up. 

For pull of a flat image before a pull of a stacked image will not reuse flat `disk.img`, since we use APFS clone we can in fact reuse the `disk.img` by just creating a clone of the already pulled flat image into the content store and not needing to re-pull.

Sources/tart/Commands/Clone.swift
edi-oai: I'm fine with `--stacked` too.

We can later add `--flat` to `tart clone`, allowing users to create a flat clone from a stacked VM image.

edi-oai (dismissed):

Sources/tart/VMDirectory+OCI.swift
```diff
@@ -137,7 +198,151 @@ extension VMDirectory {
...
+        group,
+        contentStore: contentStore,
+        registry: registry,
+        chunkSizeMb: chunkSizeMb,
+        concurrency: concurrency
+      ))
+    }
+
+    let frozenOverlayURL = try Config().tartTmpDir.appendingPathComponent("\(UUID().uuidString).asif")
+    try FileManager.default.copyItem(at: overlayURL, to: frozenOverlayURL)
```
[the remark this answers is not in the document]
yzhuang-oai: Fixed.

edi-oai (approved):

PR #1310, opened 2026-08-12
Hash files in chunks

edi-oai:
Noticed high RAM usage while running `tart clone --stacked`.


edi-oai:
Closing because this seems to be already present in https://github.com/openai/tart/pull/1307, but with `autoreleasepool { ... }`.

PR #1311, opened 2026-08-14
Fail VM startup when its control socket cannot bind

yzhuang-oai:
Bind the VM control socket before continuing startup. If binding fails, fail the VM instead of leaving it running without guest execution support.

The existing background socket lifecycle remains unchanged after a successful bind.

Testing: all 101 Swift tests pass (four skipped), including new successful-bind and bind-failure regression tests.

### DIFF
--- a/Sources/tart/ControlSocket.swift
+++ b/Sources/tart/ControlSocket.swift
@@ -6,17 +6,20 @@ import NIOPosix
 
 @available(macOS 14, *)
 class ControlSocket {
+  typealias ServerChannel = NIOAsyncChannel<NIOAsyncChannel<ByteBuffer, ByteBuffer>, Never>
+
   let controlSocketURL: URL
   let vmPort: UInt32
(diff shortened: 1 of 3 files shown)


Sources/tart/ControlSocket.swift
```diff
@@ -16,7 +18,7 @@ class ControlSocket {
     self.vmPort = vmPort
   }
 
-  func run() async throws {
+  func bind() async throws -> ServerChannel {
```
edi-oai: Could we simplify this by doing everything in `init() async throws` and remove the separate `bind()` step?

We can add another initializer later for consumers that need to provide their own `ServerChannel`.

edi-oai (dismissed):
LGTM overall, left a small nit.

Sources/tart/ControlSocket.swift
yzhuang-oai: Sounds good, fixed.

fkorotkov-oai (approved):

PR #1321, opened 2026-08-28
Preserve running VM directories during replacement

yzhuang-oai:
Cloning over a running VM can unlink its files while the process keeps consuming a VM slot. Tart loses track of it, and replacement VMs can hit the system limit.

Reject replacement when the destination's PID lock is held. Add storage locking to create and rename so they cannot race VM startup.

### DIFF
--- a/Sources/tart/Commands/Rename.swift
+++ b/Sources/tart/Commands/Rename.swift
@@ -18,6 +18,9 @@ struct Rename: AsyncParsableCommand {
 
   func run() async throws {
     let localStorage = try VMStorageLocal()
+    let lock = try FileLock(lockURL: Config().tartHomeDir)
+    try lock.lock()
+    defer { withExtendedLifetime(lock) {} }
 
     if !localStorage.exists(name) {
       throw ValidationError("failed to rename a non-existent local VM: \(name)")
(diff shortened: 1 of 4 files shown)


Sources/tart/Commands/Rename.swift
```diff
@@ -18,6 +18,9 @@ struct Rename: AsyncParsableCommand {
 
   func run() async throws {
     let localStorage = try VMStorageLocal()
+    let lock = try FileLock(lockURL: Config().tartHomeDir)
+    try lock.lock()
+    defer { withExtendedLifetime(lock) {} }
```
fkorotkov-oai, line 23: Let's move this check after validations right before `try localStorage.rename(name, newName)`

fkorotkov-oai (approved):
LGTM, left a small nitpick.

Sources/tart/Commands/Rename.swift
yzhuang-oai: These checks read mutable filesystem state, so I think they need to stay under the lock. Otherwise two concurrent renames can both see an absent destination, and the second can overwrite the first after acquiring the lock. The PID lock only rejects running destinations.

sylvesterkaczmarek (commented):
Checked the lock ordering here. Create/rename now take the same storage lock that `run` holds until the VM PID lock is established, and `replace` checks that destination PID lock before unlinking anything. That closes the lookup/publish race this PR is targeting. The focused integration coverage matches that path and CI is green.

PR #1324, opened 2026-08-31
Run ErrorReportingTask operations on the caller's actor

samestep:
This PR restores the ability #842 added to use SIGUSR2 to gracefully stop a VM, which was accidentally broken by #1262.

### DIFF
--- a/Sources/tart/Utils.swift
+++ b/Sources/tart/Utils.swift
@@ -6,8 +6,11 @@ import Foundation
 struct ErrorReportingTask {
   let task: Task<Void, Never>
 
+  // Inherit the caller's actor context, exactly as Task.init does. Without this, an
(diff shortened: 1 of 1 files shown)


sylvesterkaczmarek (commented):
This looks like the right fix for the SIGUSR2 regression. `ErrorReportingTask` wraps `Task { ... }`, so the closure parameter needs to inherit the caller actor for the wrapper to behave like direct `Task.init`; that keeps the `requestStop()` path on the expected actor without changing the task/error handling.

yzhuang-oai:
@samestep can you rebase? we needed to fix CI yesterday

samestep:
Rebased!

yzhuang-oai (approved):

PR #1327, opened 2026-09-03
Support getting the MAC address, and setting an arbitrary MAC address

mrpippy:


sylvesterkaczmarek (commented):
`VZMACAddress(string:)` validates the address format, but it can still represent multicast or broadcast addresses. `set --mac ff:ff:ff:ff:ff:ff` can therefore be accepted and persisted even though it is not a usable interface MAC. Could this require `isUnicastAddress` before saving the config?

mrpippy:
> `VZMACAddress(string:)` validates the address format, but it can still represent multicast or broadcast addresses. `set --mac ff:ff:ff:ff:ff:ff` can therefore be accepted and persisted even though it is not a usable interface MAC. Could this require `isUnicastAddress` before saving the config?

Great point, fixed.

sylvesterkaczmarek (approved):
The `--mac` path now requires `VZMACAddress.isUnicastAddress` before persisting the address, so multicast/broadcast values are rejected. This resolves my review finding.

PR #1224, opened 2026-04-07
fix(control-socket): use .path instead of .path() for stale socket cleanup

smcloughlinTT:
### Fix

 Change .path() → .path so removeItem receives the absolute path and can actually find and delete the stale socket.

 ### Testing

 - Confirmed the bug reproduces on macOS Sequoia 15.7.3 with Tart 2.32.0 using both Ubuntu and Rocky 9 VMs
 - Verified the fix resolves the issue: built patched Tart, reproduced the stop/start cycle, and tart exec works correctly after restart
 - Added unit tests verifying controlSocketURL resolves correctly for both the absolute cleanup path and the relative binding path

 ### Manual Verification

 Tested on macOS Sequoia 15.7.3 with a Rocky Linux 9 aarch64 VM (ghcr.io/cirruslabs/rocky:latest).

 Test A — Reproduce the bug with stock Tart 2.32.0:

 ```bash
# Remove any leftover socket for a clean starting state
rm -f ~/.tart/vms/rocky9/control.sock

# Start VM and exec -- works fine on first run
tart run rocky9 --no-graphics &
tart exec rocky9 uname -a   # succeeds

# Stop and restart
tart stop rocky9
ls ~/.tart/vms/rocky9/control.sock   # stale socket present
tart run rocky9 --no-graphics &

# Second exec after stop/start -- hangs, times out after 15s (exit 124)
timeout 15 tart exec rocky9 uname -a   # FAILS

tart stop rocky9
 ```

 Test B — Verify the fix with patched Tart:

 ```bash
# Change this to your own repo path
PATCHED=~/repos/tart/.build/arm64-apple-macosx/debug/tart

# Remove any leftover socket for a clean starting state
rm -f ~/.tart/vms/rocky9/control.sock

# Start VM with patched binary and exec -- works fine on first run
$PATCHED run rocky9 --no-graphics &
tart exec rocky9 uname -a   # succeeds

# Stop and restart with patched binary
tart stop rocky9
ls ~/.tart/vms/rocky9/control.sock   # stale socket present
$PATCHED run rocky9 --no-graphics &

# Second exec after stop/start -- patched binary cleans up stale socket
timeout 15 tart exec rocky9 uname -a   # succeeds

tart stop rocky9
 ```

 Fixes: https://github.com/cirruslabs/tart/issues/1220

PR #1328, opened 2026-09-04
Fix control socket paths with special characters

### LINKED ISSUE #1319: "Address already in use" in 2.36.0
Hi,

For some time, I have been running a VM via a command like the following without issues:

`TART_HOME="/Some/Tart/Home/Path" /Applications/Tart.app/Contents/MacOS/tart run some_vm_name --net-bridged="Wi-Fi" --capture-system-keys --dir="Shared:/Some/Shared/Path"`

However, after upgrading from Tart 2.35.0 to 2.36.0, I get the following error:

`bind(descriptor:ptr:bytes:): Address already in use) (errno: 48)`

How do I resolve this?
Thank you!

Tart 2.36.0
macOS 26.6.2 (25G83)

yzhuang-oai:
`URL.path()` percent-encodes VM directory names and, on older macOS releases, can return only the relative part of a URL with a base. These behaviors can prevent stale control-socket cleanup or directory changes; in Tart 2.36.0, the resulting bind error aborts VM startup.

Remove stale sockets with `FileManager.removeItem(at:)` and use `.absoluteURL.path(percentEncoded: false)` for directory changes in `run`, `exec`, and agent IP lookup. Add one regression test for a stale socket in a long path containing spaces, `%`, `#`, and Unicode.

Fixes #1319. See also #1224 and [swiftlang/swift-foundation#1011](https://github.com/swiftlang/swift-foundation/issues/1011) for the older relative-URL behavior.

## Validation

- All 126 Swift tests passed on macOS 26.6.2.
- Manual macOS VM starts succeeded with spaces in `TART_HOME` and the VM name. Host Virtualization shutdown errors and VM limits prevented completing the remaining runtime checks.


sylvesterkaczmarek (commented):
Using the decoded `URL.path` at the filesystem boundary is the right fix here. The control socket logic operates on POSIX paths, so percent-encoding from URL presentation should never reach `chdir`, unlink, or socket-path handling.

The regression is good because it combines characters that force URL encoding with the Unix-socket path-length constraint and a stale socket entry. One small strengthening would be to assert in the fixture that the encoded URL representation actually differs from `.path`, so the test remains explicitly tied to this regression if Foundation URL behavior changes.

edi-oai:
Possibly related:

* https://github.com/openai/tart/pull/1224
* https://github.com/swiftlang/swift-foundation/issues/1011

edi-oai (dismissed):
Since `.path` is deprecated, should we just use the `.path(percentEncoded: false)` instead?

There's also [`removeItem(at:)`](https://developer.apple.com/documentation/foundation/filemanager/removeitem(at:)) that directly takes a `URL`.

yzhuang-oai:
> Since `.path` is deprecated, should we just use the `.path(percentEncoded: false)` instead?
> 
> There's also [`removeItem(at:)`](https://developer.apple.com/documentation/foundation/filemanager/removeitem(at:)) that directly takes a `URL`.

Fixed, thanks @edi-oai

edi-oai (approved):

PR #1329, opened 2026-09-04
Fix Softnet closing Tart's stdout

### LINKED ISSUE #1323: --net-softnet closes tart's own stdout, so the VNC URL is never printed
Since 2.34.0, a VM started with `--net-softnet` loses everything tart prints to stdout after the VM starts. The most visible casualty is the `VNC server is running at ...` line, which is the only place the per-boot VNC passphrase ever appears.

tart 2.36.0 (Homebrew), macOS 26.6.2, M4 Max. Same VM, twice:

```
$ tart run vncprobe --no-graphics --vnc-experimental >runA.log 2>&1 &
$ cat runA.log
VNC server is running at vnc://:<four-word-passphrase>@127.0.0.1:65111

$ tart run vncprobe --no-graphics --vnc-experimental \
    --net-softnet --net-softnet-block=0.0.0.0/0 --net-softnet-allow=@host >runB.log 2>&1 &
$ wc -c runB.log
0 runB.log
```

The second VM is up and its VNC server is listening, so only the URL is gone:

```
$ lsof -nP -p 6950 -a -iTCP -sTCP:LISTEN
tart    6950 ahegyes   12u  IPv6 ...  TCP *:65112 (LISTEN)

$ lsof -p 6950 -a -d 1
tart    6950 ahegyes    1u  KQUEUE   count=0, state=0xa
```

fd 1 is a kqueue because `Softnet.run()` closed it:

```swift
// Sources/tart/Network/Softnet.swift:98
defer { try? (process.standardOutput as? FileHandle)?.close() }
```

`process.standardOutput` is only assigned when `--net-softnet-control-fd` was passed. Without it the property was never set, and Foundation resolves an unset `standardOutput` to the parent's own stdout (`_NSStdIOFileHandle`), so the defer closes tart's fd 1 rather than a softnet-owned descriptor. The line arrived in #1287 and is unchanged in 2.34.0, 2.35.0, 2.36.0 and main; 2.33.1 has no defer and prints the URL fine under softnet.

`network.run()` is the first statement of `VM.start()` (`VM.swift:248`), so the close lands before every run-time print, not just the VNC one: suspend and snapshot progress and `Requesting guest OS to stop...` go the same way. Swift's `print()` discards the EBADF, so none of it is diagnosable from the output. It lands hardest on CI, since the `noGraphics || CI` branch at `Run.swift:560` is the one that prints the URL instead of opening it, and CI is where softnet is meant to be used.

Closing the retained control handle instead of reading back `process.standardOutput` would keep the intended close and drop the accidental one. `controlFileHandle` is a local in `init()` today, so it would need to become a property.

Found this in a wrapper of mine that boots VMs under a softnet policy and has to read the URL back; there is no other way to get the passphrase, as far as I can tell.

yzhuang-oai:
Running with `--net-softnet` without a control FD closed Tart's inherited stdout, hiding the VNC URL and subsequent output. Retain the optional control handle and close only that handle after attempting to launch Softnet.

Fixes #1323.

## Validation

- All 11 existing `SoftnetControlFDTests` passed.
- A temporary regression test verified stdout survives successful and failed launches without a control FD.
- SwiftFormat lint and `git diff --check` passed.

### DIFF
--- a/Sources/tart/Network/Softnet.swift
+++ b/Sources/tart/Network/Softnet.swift
@@ -11,14 +11,13 @@ enum SoftnetError: Error {
 
 class Softnet: Network {
   private let process = Process()
+  private var controlFileHandle: FileHandle?
   private var monitorTask: Task<Void, Error>? = nil
   private let monitorTaskFinished = ManagedAtomic<Bool>(false)
 
   let vmFD: Int32
 
   init(vmMACAddress: String, extraArguments: [String] = [], controlFD: Int32? = nil) throws {
-    var controlFileHandle: FileHandle?
-
     if let controlFD = controlFD {
       guard controlFD > STDERR_FILENO else {
         throw SoftnetError.InitializationFailed(why: "Softnet control file descriptor must be greater than 2")
@@ -95,7 +94,7 @@ class Softnet: Network {
   }
 
   func run(_ sema: AsyncSemaphore) throws {
-    defer { try? (process.standardOutput as? FileHandle)?.close() }
+    defer { try? controlFileHandle?.close() }
 
     try process.run()
 


sylvesterkaczmarek (commented):
The ownership change is the right shape. `Process.standardOutput` can point at the process-wide `FileHandle.standardOutput`, so closing whatever happens to be assigned there makes `Softnet` own a descriptor it did not create. Keeping the wrapped control fd in a separate optional handle makes that ownership explicit and lets `run()` close only its own resource.

The regression that writes to stdout after `run()` is particularly useful because it catches the caller-visible breakage rather than only checking object state. I do not see another issue in this path.

edi-oai (approved):

PR #1334, opened 2026-09-16
Add --dry-run to tart prune

### LINKED ISSUE #1295: Add --dry-run to tart prune
## Problem

`tart prune` deletes matching caches or VMs immediately. There’s no way to preview what will be removed.

## Proposal

Add `--dry-run`:

```console
tart prune --entries caches --older-than 7 --dry-run
```

It should use the same selection logic, make no changes, and show matching entries plus total estimated space reclaimed.

## Motivation

Pruning has previously selected more data than intended (#271, #549). A preview would help users verify destructive cleanup before running it.

akim136:
`tart prune` deletes straight away, with no way to check what it's about to remove first.
Since it has twice taken more than intended (#271, #549), a preview seemed worth having.

`--dry-run` lists what would go, in the order it would go, with sizes and a total, and
touches nothing on disk.

```console
$ tart prune --entries caches --space-budget 10 --dry-run
Dry run: no changes will be made.
Would remove /Users/me/.tart/cache/.../ghcr.io/org/image/tag (4.2 GB)
Would remove /Users/me/.tart/cache/IPSWs/UniversalMac_15.2_24C101.ipsw (8 KB)
Total estimated space reclaimed: 4.2 GB (allocated size).
```

### How it works

Preview and real runs share one selection path: a `PruningOperation` passed into
`pruneOlderThan` and `pruneSpaceBudget` either deletes or records what it picked. A separate
preview implementation would drift from the real one and end up lying, so there isn't one.

Two things needed care.

**The budget loop only makes progress because each `delete()` shrinks the list it rebuilds.**
Take the delete away and it never finishes. Preview mode therefore remembers what it picked
and feeds that back in on the next rebuild.

**Simply filtering those out afterwards isn't enough.** Deleting a stacked cache record hands
its shared content to another record, pushing that record's charged size up — which can take
it over budget, so a real run deletes it too. Ownership is computed from the records on disk,
so a preview that filtered its own results never saw this and could list *fewer* entries than
the real run deletes. That's the dangerous direction for a preview, so the simulated removals
go into the storage instead, via `PrunableStorage.prunables(simulatingRemovalOf:)`, and
ownership and content-reference scans recompute exactly as they would after a real delete.
The protocol default filters by URL, which is right for IPSW and local VM storage;
`VMStorageOCI` overrides it. Real runs pass an empty set and behave as before.

`PruneTests` and the existing stacked-content test both pin this: preview and real deletion
must select the same records in the same order.

### Reading the numbers

Per-entry sizes are allocated usage when each entry is selected; the total is the net drop in
usage across the simulation. So content that only changes owner isn't double-counted, content
swept by garbage collection is, and where images share content the entries can add up to more
than the total. Sizes use `allocatedSizeBytes()`, since `sizeBytes()` overstates sparse disk
images — either way an estimate, not a promise of free space.

### Keeping the dry run dry

Three paths would otherwise have written to disk: `--gc`, the garbage collection `Root` runs
before *every* command (so a dry run mutated state before `Prune.run()` was even reached), and
building the storages, since `Config.validateTartHome()` creates directories despite its name.
The first two are skipped, the storages take an optional `readOnly` flag, and existing callers
plus automatic low-disk pruning are untouched.

A fourth was subtler and came out of review. `VMStorageOCI.list()` recursively enumerated
each cached-image record, and on filesystems that update directory atimes during enumeration
that advances the very timestamp age and LRU selection read back — so a preview could make old
entries look recently accessed and change what the real prune then removed. Cache records are
leaves in the namespace and their contents are already addressed by known path, so the scan now
skips their descendants. Record selection timestamps are preserved exactly, to the nanosecond,
with no restoration writes; the real scan stops paying for the traversal too. Enumerated
ancestor directories and manifest files still acquire atimes, but those are not selection
inputs.

Those skips are the only things a dry run doesn't model, and it says so after the results when
they apply: `--gc` can remove untagged records the preview didn't list, and temporary cleanup
can release content pinned by stale temporary manifests, pushing more images over budget.

### Tests and checks

New `PruneTests.swift` covers both criteria, ordering, termination, shared ownership, content
GC, failure paths and that a real run still deletes. The existing stacked-image test now asserts
preview and deletion agree rather than differ. Its fixtures originally set *future* access times
so enumeration could not disturb their ordering — which encoded the atime defect above as a
workaround — and now use explicit past ordering and assert the timestamps are untouched.

`swift build` and `swift test` pass (143 tests, 4 skipped, 0 failures) and SwiftFormat needs no
changes, run against a throwaway `TART_HOME` with `/usr/sbin` on `PATH` as CI does. The 4 skips
are the existing Docker-dependent tests skipping themselves.

I couldn't run the Go integration suite locally since it boots real VMs, so that's still
outstanding.

Closes #1295


akim136:
Good catch — this was real, and the future-atime fixtures were me papering over it rather than noticing it. Fixed in e2d446c.

I confirmed the mechanism before changing anything: on APFS, enumerating a directory advances that directory's atime (reading a file inside advances the file's, not the directory's), and `CachedImagePrunable.accessDate()` → `VMDirectory.accessDate()` reads `contentAccessDateKey` on exactly that record directory. So `list()` recursing into records was re-dating the selection inputs, and a preview could change what the real prune went on to remove.

Cache records are leaves in the namespace, and recognition, manifest reads and sizing already address their contents by known path rather than by enumeration, so `list()` now calls `skipDescendants()` once it recognises one. Record selection timestamps are preserved exactly, to the nanosecond, with no restoration writes — and the real scan stops paying for the traversal too.

To be precise about the limit of the claim: enumerated ancestor directories and manifest files still acquire atimes. Those are not selection inputs, and avoiding them would need read-without-atime support or a separate metadata index.

The fixtures now use explicit past ordering and assert the timestamps are untouched. Those assertions fail against 6349d30 with the expected signature — a record's atime coming back as "now" instead of its fixed past value.

Separately, while looking at this I think `URL.updateAccessDate()` has a pre-existing bug: it reads `.contentAccessDateKey` but passes that value to `utimes` as the modification time, so each call overwrites mtime with the old atime. It has four callers (`VMStorageLocal.swift:23`, `VMStorageOCI.swift:138`, `VM.swift:99`, `ContentStore.swift:105`). I've left it alone since it predates this PR and nothing here depends on it — happy to fix it here or file it separately, whichever you prefer.

